PR #17, opened 2016-10-16
Selinux fix so containers get build on CentOS/RHEL

rabem00:
When running a make build (or make refresh) the docker volume ${PWD}:/galaxy was not accessible because of a selinux issue on cents 7.x or rhel 7.x. Result the build fails with the following issue:

[Errno 13] Permission denied: '/galaxy/requirements.txt

A complete output of the issue:
```
ansible-container_1  | TASK [Clear yum cache] *********************************************************
ansible-container_1  | changed: [django]
ansible-container_1  |  [WARNING]: Consider using yum module rather than running yum
ansible-container_1  | 
ansible-container_1  | TASK [Install python packages] *************************************************
ansible-container_1  | fatal: [django]: FAILED! => {"changed": false, "cmd": "/usr/bin/pip install -r /galaxy/requirements.txt", "failed": true, "msg": "\n:stderr: You are using pip version 7.1.0, however version 8.1.2 is available.\nYou should consider upgrading via the 'pip install --upgrade pip' command.\nCould not open requirements file: [Errno 13] Permission denied: '/galaxy/requirements.txt'\n"}
ansible-container_1  |  [WARNING]: Could not create retry file 'main.retry'.         [Errno 2] No such
ansible-container_1  | file or directory: ''
ansible-container_1  | 
```

The docker inspect command shows:  /galaxy:rw , connecting with _make shell_ and trying to access the /galaxy directory will result in a permission denied. This could be fixed by adding option behind the /galaxy volume. This works but the next selinux problem occurs:

```
ansible-container_1  | TASK [Create the /etc/galaxy directory] ****************************************
ansible-container_1  | fatal: [django]: FAILED! => {"changed": false, "failed": true, "gid": 0, "group": "root", "mode": "0755", "msg": "chgrp failed", "owner": "root", "path": "/etc/galaxy", "size": 6, "state": "directory", "uid": 0}
ansible-container_1  |  [WARNING]: Could not create retry file 'main.retry'.         [Errno 2] No such
ansible-container_1  | file or directory: ''
```

So we need another fix then the z option described in [see volumes selinux for more info](http://www.projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/)

**By adding more privileges in the container.yml file everything gets build without problems. So adding privileged: yes appears to be the solution.**

My system is up to date and running:
CentOS Linux release 7.2.1511 (Core) 
No packages marked for update

### DIFF
--- a/ansible/container.yml
+++ b/ansible/container.yml
@@ -3,6 +3,7 @@ services:
   django:
     image: centos:7
     working_dir: /galaxy
+    privileged: yes
     environment:
       - PGPASSWORD={{ galaxy_postgres_password }}
     volumes:
@@ -25,7 +26,7 @@ services:
     environment:
       - POSTGRES_PASSWORD={{ galaxy_postgres_password }}
       - POSTGRES_USER={{ galaxy_postgres_user }}
-      - POSTGRES_DB={{ galaxy_postgres_db }} 
+      - POSTGRES_DB={{ galaxy_postgres_db }}
 
   elastic:
     image: elasticsearch:latest


chouseknecht:
Thanks for trying out Galaxy and working on this issue. Your time and contribution are appreciated. 

I changed all the volumes to have ':Z' permission, which is maybe slightly better than running in privileged mode, and it seems to work. Can you please give that a try.

I don't want to run with `privileged: true`, because that generally sets off alarm bells for people. I also don't understand the point of having SELinux enabled, if you're then going to run containers in privileged mode. If that's the case, then why not just disable SELinux? This seems to be a recurring theme for those developing in an SELinux environment.

rabem00:
I will verify by doing a clean install of the system and galaxy. (On disable of SELinux, when enabled but with more permissions you still get the audit warnings). In pull request 16 I did the same as you're trying now. I closed that one and went for  privileged option. You'll here from me later on.

rabem00:
I did a clean install and cloned the fix you made, but that doesn't work. I get a different error:
```
ansible-container_1  | PLAY [django] ******************************************************************
ansible-container_1  | 
ansible-container_1  | TASK [Download postgres repo package] ******************************************
ansible-container_1  | fatal: [django]: UNREACHABLE! => {"changed": false, "msg": "Authentication or permission failure. In some cases, you may have been able to authenticate and did not have permissions on the remote directory. Consider changing the remote temp path in ansible.cfg to a path rooted in \"/tmp\". Failed command was: ( umask 77 && mkdir -p \"` echo $HOME/.ansible/tmp/ansible-tmp-1476640411.34-91323859000389 `\" && echo ansible-tmp-1476640411.34-91323859000389=\"` echo $HOME/.ansible/tmp/ansible-tmp-1476640411.34-91323859000389 `\" ), exited with result 1", "unreachable": true}
ansible-container_1  |  [WARNING]: Could not create retry file 'main.retry'.         [Errno 2] No such
ansible-container_1  | file or directory: ''
```

If I add `privileged: yes` in the container.yml (and leave everything as you committed it) than the build goes well.

Don't know the answer right now, will do some more debugging tomorrow and I will try to come up with a different solution.

rabem00:
Found many references about the same error and as suspected a problem with permission access. (For instance the same error with possible solution can be [found here](https://davesnigier.com/changing-ansible-temporary-directory/)) But that solution is one of edit the ansible.cfg file (that I can't find in the build tree), so that's no option.

chouseknecht:
My bad. Need to use :z rather than :Z when multiple containers access the same mount point. See [docker run --volumes-from](https://docs.docker.com/engine/reference/commandline/run/#/mount-volumes-from-container---volumes-from) for the details.

rabem00:
I'm getting the same error on your last commit. I also tried a clean server install and new git clone, but still getting the same error. Also tried putting the other to :z, but without any luck. So the error still exist!

rabem00:
Will close the pull request and post a bug

PR #18, opened 2016-11-25
Note that builds can take a long time

goetzk:
Because it isn't immediately obvious to someone who isn't familiar
with Docker I think its worth mentioning that building will take some
time.


chouseknecht:
@goetzk, thanks for taking the time to work on this, and the feedback is appreciated.

PR #19, opened 2017-01-06
Support meta file with *.yaml file extension

jtyr:
This patch is adding support for meta file with `*.yaml` file extension. This is quite important thing as Ansible supports both file extensions (`yml` and `yaml`) and only Galaxy doesn't.

### DIFF
--- a/galaxy/main/celerytasks/tasks.py
+++ b/galaxy/main/celerytasks/tasks.py
@@ -74,16 +74,24 @@ def get_repo_raw(token, repo_name):
 
 
 def get_meta_data(repo):
+    '''
+    Try to load the meta file from the repo.
+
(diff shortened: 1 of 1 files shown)


chouseknecht:
@jtyr- 

Nice! Thank your for looking at this. I will test it, and hopefully get it merged today. 

--Chris

jtyr:
Thanks for merging it, @chouseknecht. When do you think it might be deployed on the Galaxy portal?

alikins:
Should probably add a galaxy side import rule to warn about main.yaml and suggest main.yml while we are at it.

And possibly throw an error if both are included. (Mentioned at https://github.com/ansible/ansible/pull/46505#issuecomment-436324139)

PR #20, opened 2017-02-01
Using copy instead of synchronize

roolo:
Synchronize causes proble while build this project

~~~
ansible-container_1  | TASK [django-role : Copy source] ***********************************************
ansible-container_1  | fatal: [django]: FAILED! => {"changed": false, "failed": true, "msg": "synchronize uses rsync to function. rsync needs to connect to the remote host via ssh or a direct filesystem copy. This remote host is being accessed via docker instead so it cannot work."}
~~~

### DIFF
--- a/ansible/roles/django-role/tasks/main.yml
+++ b/ansible/roles/django-role/tasks/main.yml
@@ -66,7 +66,7 @@
     - /venv
 
 - name: Copy source
-  synchronize:
+  copy:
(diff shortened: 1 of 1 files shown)


chouseknecht:
@roolo 

Apologies for not getting to this sooner.

The build has been updated to use the latest version of Ansible Container, and the synchronize module seems to be working as expected.

PR #26, opened 2017-10-04
Fix django container failure on restart

cutwater:
Use manage.py script from project directory for
executing database migrations on container restart.


chouseknecht:
Looks good to me.

PR #85, opened 2017-10-13
Django container startup optimizations

cutwater:
* Use bash scripts to perform initial database migrations in django
container during startup process.
* Reduce startup speed by removing import of test data.
* Do not refresh galaxy indexes on startup.


chouseknecht:
I'm able to run `make build` (from scratch, no prior images) and `make run`.

LGTM. Merging.

Issue #82, opened 2017-10-13
Use SPDX license identifiers


chouseknecht:
_From @ypid on July 17, 2016 9:7_

SPDX license identifiers seem to have become the default license identifier format AFAIK. I think it would make sense to use them also for `meta/main.yml`?

As of Ansible 2.1 `ansible-galaxy init` generates:

``` YAML
  # Some suggested licenses:
  # - BSD (default)
  # - MIT
  # - GPLv2
  # - GPLv3
  # - Apache
  # - CC-BY
  license: license (GPLv2, CC-BY, etc)
```

With this proposal implemented, this could look like:

``` YAML
  # Some suggested licenses:
  # - BSD-2-Clause (default)
  # - MIT
  # - GPL-2.0
  # - GPL-3.0
  # - AGPL-3.0
  # - GPL-3.0+
  # - Apache-2.0
  # - CC-BY-SA-4.0
  license: AGPL-3.0
```

The current format is ambiguous to say the least.

Ref: https://spdx.org/licenses/

_Copied from original issue: ansible/galaxy-issues#169_

cutwater:
Addressed by #89

pombredanne:
FWIW, the use of a list of licenses in documentation comments has some funny side effects when using a license scanner: https://github.com/nexB/scancode-toolkit/issues/1111 (and ScanCode is also used in openshift afaik) 

I  anycase I second using SPDX identifiers (and license expressions) in general (I am one of the SPDX co-founders ;) )

Issue #58, opened 2017-10-13
Inconsistent licensing info in ansible/galaxy repo


chouseknecht:
_From @grossws on February 7, 2017 21:40_

https://github.com/ansible/galaxy repo contains `LICENSE` file with Apache License v2 and `license='Proprietary'` + `'License :: Other/Proprietary License',` in `classifiers` in `setup.py`.

_Copied from original issue: ansible/galaxy-issues#233_

chouseknecht:
Fixed.

grossws:
@chouseknecht, also `package.json` has `"license": "Commercial"` in it

PR #89, opened 2017-10-14
Fix license identifiers in project metadata

cutwater:
* Use a valid SPDX license identifier in package.json and setup.py
  files, refer to https://spdx.org/licenses/
* Use a valid PyPi license classifier, refer to
  https://pypi.python.org/pypi?%3Aaction=list_classifiers
  for more details.
* Set 'private' property in package.json

Issue #58

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "galaxy",
   "version": "0.0.0",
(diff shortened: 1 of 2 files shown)


cutwater:
Closes #82

PR #90, opened 2017-10-14
Replace npm with yarn

cutwater:
Replacing npm with yarn allows dramatically speedup
nodejs dependencies installation in Docker container running on
Linux host and therefore reduce gulp container image build time.


cutwater:
Looks like CI build speed is not affected. By some reasons npm freezes during installation on some systems when running inside docker container.

Issue #91, opened 2017-10-16
RFC: Release process


cutwater:
## Problem description

Galaxy project should have a process to build and distribute release artifacts in a form of Docker container(s) published at Docker Hub.

### Action items:

- [x] Split configuration depending on environment profile (e.g. development, testing, production).
- [x] Build docker container scripts implementation.
- [x] Automated release process (build and publish docker containers to Docker Hub on release).
- [x] Install galaxy to docker containers as python package (wheel or sdist).

### 1. Application configuration

Currently settings file is autogenerated by Ansible scripts and then uploaded to container. All development specific settings will be moved to a default settings file (e.g. `INSTALLED_APPS`, etc.), which is shipped with the package.

Environment specific settings should be passed via environment variables or \ and system settings file (`/etc/galaxy/settings.py`).

### 2. Build python distribution (sdist, wheel)

Use sdist or wheel format as a build artifact to install in release docker container.

### 3. Docker container

`ansible/galaxy` - container which runs galaxy web application or celery working depending on arguments

To run web application:

```
docker run -d ansible/galaxy run web
```

To run worker:

```
docker run -d ansible/galaxy run worker
```

### 4. Static files

Static files should be included in `ansible/galaxy-web` docker web and served by gunicorn, so web application will be self-sufficient and operational in minimal installation scenario.

To setup nginx, volume containing static files may be mounted to nginx container or nginx may simply be used as a caching reverse-proxy.

Optionally static may also be release in a standalone data container?

### 5. Documentation

Installation guide should available on GitHub Wiki or Galaxy web-site?

chouseknecht:
Most of this work is already done or in progress. And it's looking really nice! Awesome work @cutwater!

PR #103, opened 2017-10-26
Serve static in production mode

cutwater:
This change is requried to deliver a Galaxy application as a
standalone container, which does not requre by default any
additional web server (e.g. nginx) to serve static, so it can
run autonomously.


cutwater:
Implements #91

PR #105, opened 2017-10-26
Add tools for building release container

cutwater:
Add scripts that build a "build" container, which is used
to compile static files and a "release" container.


cutwater:
Implements #91

PR #113, opened 2017-10-27
Load correct settings

chouseknecht:
The `galaxy-manage` command isn't loading the settings files from /etc/galaxy.


cchurch:
@chouseknecht Take a look at http://django-split-settings.readthedocs.io/en/latest/ -- it's what we use in Tower/AWX to load required/optional settings from `/etc/awx/`.

cutwater:
@cchurch, basically we implemented `include_settings` function similar to `django-split-settings` without an additional dependency. Optional settings are loaded from `/etc/galaxy` for production settings.

PR #118, opened 2017-10-27
Installer

chouseknecht:
Building an installer modeled after the AWX installer. 
- adds an `installer` directory
- adds `install.yml` playbook
- adds `inventory` file with variable for determining whether or not to perform an image build and which install path to follow
- aims to support Docker and OpenShift


installer/roles/image_build/tasks/copy_branding.yml
```diff
@@ -0,0 +1,33 @@
+# copy_branding.yml
+---
+- name: Copy templates
+  copy:
+    src: '{{ item }}'
+    dest: '{{ playbook_dir }}/../galaxy/templates/'
```
cutwater: Better to use same approach as in main.yml for `root_path`

```
- name: Set fact root_path 
   set_fact: 
     root_path: '{{ playbook_dir }}/..' 
   tags:
   - start
   - stop
```

installer/roles/image_build/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
+---
+- name: Set fact root_path 
+  set_fact: 
+    root_path: '{{ playbook_dir }}/..' 
+  tags:
+  - start
+  - stop
+
+- name: Get version from checkout if not provided
+  shell: git describe --long | sed 's/-g.*//' | sed 's/-/./'
```
cutwater: Is playbook going to be executed against develop \ master branches only or it can be executed for any commit \ pull request? In second case version transformation from `v2.1.2-169-gbeae8f0` to `v2.1.2.169` is not reliable. In this example value `169` indicates offset (nuber of commits) from latest tag. In case of two branches with similar offset command above will return the same result. For example:

For git history that has 2 branches started from develop and each branch contains one commit:

```
* 9527d18 (fix/logging) Do not write production logs to files
| * beae8f0 (fix/release-entrypoint) Fix tini execution from PID != 1
|/  
*   f9125bb (develop) Merge pull request #124 from cutwater/feature/publish-dev-container
```

`git describe` after transformation above will return similar results for different commits, which might be unexpected behavior:

```
$ git checkout fix/release-entrypoint
Switched to branch 'fix/release-entrypoint'
$ git rev-parse HEAD
beae8f024ab54d9dfbbabbf602bbf8c5f7506b2e
$ git describe --tags --long | sed 's/-g.*//' | sed 's/-/./'
v2.1.2.169
$ git describe --tags --long                                
v2.1.2-169-gbeae8f0

$ git checkout fix/logging
Switched to branch 'fix/logging'
$ git rev-parse HEAD        
9527d189828ecd181a7500026d027982130359f7
$ git describe --tags --long | sed 's/-g.*//' | sed 's/-/./'
v2.1.2.169
$ git describe --tags --long                                
v2.1.2-169-g9527d18
```

chouseknecht: I don't know. It's not that important. This isn't intended for use during development, so I think the commit count is perfectly fine.

PR #126, opened 2017-10-31
Do not write production logs to files

cutwater:


cutwater:
Created as a standalone commit to fix broken production settings.

PR #136, opened 2017-11-03
Ignore W503 - an invalid flake8 warning

cutwater:
W503 is an outdated warning which doesn't reflect actual PEP8[1] version
after update[2].
Will be replaced in future pycodestyle release[3].

[1] https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator
[2] https://hg.python.org/peps/rev/3857909d7956
[3] PyCQA/pycodestyle#498

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,5 +6,7 @@
 # E501 - Line too long
 # E722 - Do not use bare except
 # E731 - Prefer def over lambda
-ignore=E221,E231,E241,E251,E265,E501,E722,E731
+# W503 - line break before binary operator, to be replaced with W504.
+#        Refer to https://github.com/PyCQA/pycodestyle/issues/498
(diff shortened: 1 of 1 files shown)


juanvalino:
I'm having this problem too. Any fix or workaround?

cutwater:
@juanvalino I don't think I understood your issue. As I checked recently it seems that `W503` is enabled in flake8 by default. The only solution so far is to ignore it in config file.

juanvalino:
@cutwater finally I solved the problem adding a `# noqa: W503` comment to the file, as seems that galaxy server ignores config files.

cutwater:
@juanvalino This patch was related to flake8 config that we're using to check galaxy codebase itself. Linter that validates roles uses standalone config.

Thanks for pointing that out, it's a bug. I submitted issue #816 to track it.

PR #179, opened 2017-11-13
Add django management command 'show_urls'

cutwater:
Add a django management command 'show_urls' for development
purposes. It displays allconfigured URL patterns and
corresponding view callables.

### DIFF
--- a/galaxy/main/management/commands/show_urls.py
+++ b/galaxy/main/management/commands/show_urls.py
@@ -1,0 +1,41 @@
+# (c) 2012-2016, Ansible by Red Hat
+#
+# This file is part of Ansible Galaxy
+#
+# Ansible Galaxy is free software: you can redistribute it and/or modify
+# it under the terms of the Apache License as published by
+# the Apache Software Foundation, either version 2 of the License, or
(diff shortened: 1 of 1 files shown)


cchurch:
Install `django-extensions` and you get the same thing: https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/commands/show_urls.py

cutwater:
@cchurch Didn't want to bring additional dependency to project for just one function

chouseknecht:
@cutwater 

I think @cchurch is right. Seems like having `django-extensions` installed during development would be really handy. It provides several useful utilities for troubleshooting things. 

Can we add it to our [dev Dockerfile](https://github.com/ansible/galaxy/blob/develop/scripts/docker-dev/Dockerfile), and then add it to installed apps in [settings/development.py](https://github.com/ansible/galaxy/blob/develop/galaxy/settings/development.py)?

cutwater:
@chouseknecht as immediate solution we can add it to dev Dockerfile, but later I'd prefer to have in in `requirements-dev.txt`.

Closing this one.

PR #261, opened 2017-12-14
Move github attributes to Repository model

cutwater:


cutwater:
Looks like we should also change `Stargazers` model refernce from `Role` to `Repository`.

PR #263, opened 2017-12-14
Build script update

corrjo:
VENV_BIN wasn't set to the bin folder just the venv folder


chouseknecht:
@corrjo Thanks for catching this!

PR #275, opened 2018-01-05
UI for Namespaces

chouseknecht:
- Adds new API endpoints for /namespaces and /providers
- Adds UI for viewing and adding user specific Namespaces
- Adds My Namespaces menu option

Still to do... 
- [x] Add Namespace edit/modify
- [x] Show a warning before deleting a namespace. Before deleting we should let the user know if associated Provider Namespaces exist, and what will happen, if they continue.
- [x] If a Namespace is deleted, we won't delete the associated Provider Namespaces, and related content. In other words, the delete should not cascade. Instead, Provider Namespaces will be disassociated, and available to associate later with another Namespace.


chouseknecht:
[Namespace refactor](https://trello.com/c/veDsPZhi/6-namespace-refactor)

galaxy/api/githubapi.py
```diff
@@ -0,0 +1,96 @@
...
+# Ansible Galaxy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# Apache License for more details.
+#
+# You should have received a copy of the Apache License
+# along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+
+from github import Github
```
cutwater: Usually it's better to use module imports over object imports
https://docs.openstack.org/hacking/latest/user/hacking.html#imports

galaxy/api/githubapi.py
```diff
@@ -0,0 +1,96 @@
...
+                    'avatar_url': org.avatar_url,
+                    'location': org.location,
+                    'company': org.company,
+                    'email': org.email,
+                    'html_url': org.blog,
+                    'followers': org.followers
+                }
+                result.append(source)
+        except GithubException, exc:
+            raise Exception(
```
cutwater: Please do not raise base `Exception` class. Let's create custom Galaxy exceptions.

PR #283, opened 2018-01-16
 Multiple repositories support

cutwater:


chouseknecht:
Looks good to me.

I see you're removing the `original_name` field from models where we don't use it. Makes sense. Did you remove `original_name` from `CommonModel` and `CommonModelNameNotUnique`? Most of the models subclass one or the other.

chouseknecht:
Looks like there is some overlap with #303, as both contain migration 0071 for removing unused `original_name` attribute. 

Once you sort that out, this looks good to me.

PR #303, opened 2018-01-19
Remove dead code and improve code quality

cutwater:


chouseknecht:
Looks like there is some overlap with #283, as both contain migration 0071 for removing unused `original_name` attribute. 

Once you sort that out, this looks good to me.

cutwater:
@chouseknecht I belive migration `0071` is displayed both in this PR and in #283 because #283 is based on this commit. Unfortunately GitHub doesn't recognize when one PR depends on another one and shows diff with base branch.

PR #317, opened 2018-02-05
Vendor javascript packages

chouseknecht:
Rather than load third party JS from `galaxy/static/bower_components` and `galaxy/static/node_modules`, going back to pre-installing or vendor-ing packages to `galaxy/static/vendor`.

The point of this change is to avoid the unreliability of installing these packages during the production build process, and insures we know exactly what we're serving from the `static` directory.

Also, removing calls to `dotdotdot.js`. This package is no FOSS, but instead requires purchasing a license.

### DIFF
--- a/galaxy/templates/header.html
+++ b/galaxy/templates/header.html
@@ -17,79 +17,81 @@
 
     <!-- CSS -->
     <!-- PatternFly Styles -->
-    <link rel="stylesheet" type="text/css" href="{% static 'node_modules/patternfly/dist/css/patternfly.min.css' %}" />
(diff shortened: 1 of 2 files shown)


cutwater:
This patch affects 808 files and it's hard to distinguish logical parts from  it. Could you please split it into smaller patches, each serving one particular purpose?

chouseknecht:
@cutwater 

OK. Let me see what I can do... I think the 800 files is mostly because I deleted the `node_modules` and `bower_components` directories.

chouseknecht:
@cutwater 

Down to 2 files. This PR relies on #323 and #325.

PR #320, opened 2018-02-06
Modernize Python 2 code to get ready for Python 3

cclauss:
Make the minimal, safe changes required to convert the code to be syntax compatible with both Python 2 and Python 3. There would definitely be more work required to complete the port of the code to Python 3 but this is a minimal, safe first step.

    Run: futurize --stage1 -w **/*.py

See Stage 1: "safe" fixes http://python-future.org/automatic_conversion.html#stage-1-safe-fixes

$ pip install future
$ git checkout -b modernize-python2-code
$ futurize --stage1 -w **/*.py # done in zsh for ** globbing
$ git commit --all -m "Modernize Python 2 code to get ready for Python 3"
$ git push --set-upstream origin modernize-python2-code

### DIFF
--- a/deploy/ec2/ec2.py
+++ b/deploy/ec2/ec2.py
@@ -127,6 +127,7 @@ destination_format_tags: Name,clusterid,deployment,private_dns_name
 These settings would produce a destination_format as the following:
 'webserver-ansible-blue-172.31.0.1'
 '''
+from __future__ import print_function
 
 # (c) 2012, Peter Sankauskas
 #
--- a/deploy/ec2/ec2.py
+++ b/deploy/ec2/ec2.py
@@ -127,7 +127,6 @@ destination_format_tags: Name,clusterid,deployment,private_dns_name
 These settings would produce a destination_format as the following:
 'webserver-ansible-blue-172.31.0.1'
 '''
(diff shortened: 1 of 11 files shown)


deploy/ec2/ec2.py
```diff
@@ -127,6 +127,7 @@
 These settings would produce a destination_format as the following:
 'webserver-ansible-blue-172.31.0.1'
 '''
+from __future__ import print_function
```
cutwater: Please move import below copyright comment.

galaxy/api/views.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
cutwater: ditto

galaxy/main/management/commands/refresh_role_counts.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
cutwater: ditto

cclauss: Done.

galaxy/api/views.py
cclauss: Done.

PR #376, opened 2018-03-16
Angular 5: non-javascript changes

chouseknecht:
All the changes made to backend, admin or other supporting files to accommodate the move to Angular 5 and a single page application (SPA) framework.

### DIFF
--- a/galaxy/api/views/account.py
+++ b/galaxy/api/views/account.py
@@ -1,0 +1,44 @@
+# (c) 2012-2018, Ansible by Red Hat
+#
+# This file is part of Ansible Galaxy
+#
+# Ansible Galaxy is free software: you can redistribute it and/or modify
+# it under the terms of the Apache License as published by
+# the Apache Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# Ansible Galaxy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# Apache License for more details.
+#
+# You should have received a copy of the Apache License
+# along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+import logging
+
+from rest_framework.status import HTTP_200_OK
+from rest_framework.authentication import SessionAuthentication
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.generics import CreateAPIView
+
+from allauth.account.adapter import get_adapter
+
+
+__all__ = (
+    'LogoutView',
+)
+
+logger = logging.getLogger(__name__)
+
+
+class LogoutView(CreateAPIView):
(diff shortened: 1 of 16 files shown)


galaxy/api/views/account.py
```diff
@@ -0,0 +1,44 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# Apache License for more details.
+#
+# You should have received a copy of the Apache License
+# along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+import logging
+
+from rest_framework.status import HTTP_200_OK
```
cutwater: It's usually recommended to import modules rather then module attributes.

galaxy/main/urls.py
```diff
@@ -25,40 +25,28 @@
...
     urlpatterns += [
-        url(r'^static/(?P<path>.*)$',
-            never_cache(serve_staticfiles))
+        url(r'^static/(?P<path>.*)$', never_cache(serve_staticfiles))
     ]
 else:
     urlpatterns += [
-        url(r'^static/(?P<path>.*)$', serve_static,
-            kwargs={'document_root': settings.STATIC_ROOT})
+        url(r'^static/(?P<path>.*)$',
```
cutwater: Can you please explain this change?

galaxy/main/views.py
```diff
@@ -129,16 +129,16 @@ def build_standard_context(request):
...
-    context = build_standard_context(request)
-    contentblocks = ContentBlock.objects.filter(name__in=[
-        'main-title',
-        'main-share',
-        'main-downloads',
-        'main-featured-blog'
-    ]).all()
-    context['contentblocks'] = {item.name: item for item in contentblocks}
-    return render_to_response('home.html', context)
+# def home(request):
```
cutwater: Please delete unneeded code.

galaxy/main/urls.py
chouseknecht: Looks like a formatting change.

chouseknecht:
Tests aren't going to pass, because the Makefile is based on the existence of `galaxyui`, and all of the new javascript. However, none of those files are included in this PR.

galaxy/main/urls.py
chouseknecht: Maybe what you're really asking about is the 2 lines appearing after your comment, which are the following:

```
    url(r'^(?P<path>.*[css|js|png|jpg|jpeg|ico|woff|woff2|svg|ttf])/?$',
        serve_static, kwargs={'document_root': settings.STATIC_ROOT}),
    url(r'^$',
        serve_static, kwargs={'document_root': settings.STATIC_ROOT, 'path': 'index.html'})
```

When `DEBUG=False`, any requests that don't match a regex pattern are likely part of the Angular app. When deployed, the Angular files (including all the enumerated file types) are deployed to `STATIC_ROOT`. However, the request will not include `/static`, so the above two patterns make sure the server try to find the requested file within the static directory.

PR #386, opened 2018-03-21
Fixes broken features of 'My Content', 'My Imports'

chouseknecht:
Required to support enable/disable of Namespaces. PR #386 expects this.
- For #388 - Added CSS transitions and `setTimeout` in JS to smoothly fade in and out spinners. Fixed absolute placement so that spinners are properly centered over related elements and content.
- For #389 - Added styling to limit flex box expansion and eliminate horizontal scrolling on My Content
- For #387 - Added proper enable/disable of buttons and menu items. Now it's clear when a Namespace is disabled. When a Namespace is disabled, the `Add Content` buttons is correctly disabled, the `Edit` menu option is disabled, and the `Disable` menu option is replaced by an `Enable` option.

### DIFF
--- a/galaxyui/src/app/app.component.html
+++ b/galaxyui/src/app/app.component.html
@@ -45,7 +45,7 @@
             </ul>
         </div>
     </pfng-vertical-navigation>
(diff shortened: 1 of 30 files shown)


chouseknecht:
@cutwater 

Fixes for broken tests in #391

PR #406, opened 2018-03-28
WIP: Improve content blocks rendering

cutwater:
* Sanitize content by using `bleach` library.
* Cache contnet in django cache.


chouseknecht:
We're consuming ContentBlocks through the API now. Does it makes sense to hook this into `api/views/content_block.py`?

cutwater:
@chouseknecht In this case let's drop everyting related to server side rendering of content blocks.
However it might be a good idea to sanitize html written to database \ returned by REST API.
Closing this PR for now.

Issue #402, opened 2018-03-26
User GitHub attributes not updating on login


chouseknecht:
We don't seem to be updating `user.github_user` and `user.github_avatar`. If you start with an empty database, and log in for the first time, these attributes are not set.

chouseknecht:
![](https://github.trello.services/images/mini-trello-icon.png) [User GitHub attributes](https://trello.com/c/dxDABvTe/44-user-github-attributes)

cutwater:
@chouseknecht That's because according to actual code c585300 we're updating `provider_namespaces`, not `accounts_customuser` table. It seems reasonable because user potentially can have multiple linked accounts and it's unclear from which account we should take avatar for user table. 
The same story is about `github_user` field, isn't it covered by `provider_namespaces` table now?

PR #426, opened 2018-04-09
Update user's github attributes on login

cutwater:
Update `github_user` and `github_avatar` fields on user login.

Issue: #402


galaxy/accounts/models.py
```diff
@@ -67,11 +69,14 @@ class CustomUser(AbstractBaseUser, PermissionsMixin, DirtyMixin):
     date_joined = models.DateTimeField(_('date joined'), default=timezone.now)
 
     karma = models.IntegerField(default=0, db_index=True)
-    github_avatar = models.CharField(_('github avatar'), max_length=254, blank=True)
-    github_user = models.CharField(_('github user'), max_length=254, blank=True)
-    cache_refreshed = models.BooleanField(_('cache refreshed'), default=False)
-
-    objects = UserManager()
+    github_avatar = models.CharField(
```
chouseknecht: I'm confused. I thought we were changing `github_user` and `github_avatar` to be properties, with the results pulled from the social account table.

cutwater: According to discussion in slack, let's rename it to `avatar` as a part of this PR.

galaxy/main/signals/handlers.py
```diff
@@ -37,21 +36,29 @@
 
 @receiver(user_logged_in)
 def user_logged_in_handler(request, user, **kwargs):
+    social = auth_models.SocialAccount.objects.get(
+        provider=constants.PROVIDER_GITHUB.lower(), user=user)
+    user.avatar_url = social.extra_data.get('avatar_url')
```
chouseknecht: This is fine for now. However, in the future, I anticipate we'll have a page where the user can manage their profile, and update their `avatar_url`. When that becomes true, we should only set the `user.avatar_url` when it is empty.

cutwater: I'm totally agree.

PR #437, opened 2018-04-09
Produce warnings for linting issues

cutwater:


chouseknecht:
Why are we changing this? If content doesn't pass lint checks, it shouldn't be imported. Otherwise, I'm not sure why we're running the lint checks.

PR #438, opened 2018-04-09
Importer fixes and tweaks

cutwater:


alikins:
>  galaxy/importer/linters.py → galaxy/importer/linters/__init__.py

Style wise, does the project lean towards a preference for empty-ish __init__.py or with code in them?  Seems to be a mix, but if pref is empty __init__ may be worthwhile to go ahead and split these into separate py modules .

galaxy/importer/linters/__init__.py
```diff
@@ -23,6 +23,9 @@
 
 LOG = logging.getLogger(__name__)
 
+LINTERS_DIR = os.path.abspath(os.path.dirname(__file__))
```
alikins: Would it be worthwhile to make this configurable? 

Don't really need it to be configurable yet though so maybe something for later.

galaxy/importer/linters/yamllint.yaml
```diff
@@ -0,0 +1,19 @@
+extends: default
```
alikins: Would it be useful to add a comment here explaining it's based on the ansible-lint rules.
(+1 for mentioning it in the commit message).

galaxy/importer/loaders/role.py
```diff
@@ -291,7 +291,7 @@ def _load_container_yml(self):
         ansible_container_meta_file = os.path.join(
             self.path, self.ANSIBLE_CONTAINER_META_FILE)
         if os.path.exists(ansible_container_meta_file):
-            if self._container_yml_type is not None:
+            if container_yml_type is not None:
                 raise exc.ContentLoadError(
```
alikins: Not directly related to this pr, but just noticed exc.ContentLoaderError is a subclass of 'ImportError' which seems wrong. Should it be a subclass of exc.ImporterError instead?

alikins:
The import tasks use 'LOG' for module scope logger instead of 'logger' or 'log'. Is 'LOG' the expected name for celery related code?

galaxy/importer/loaders/role.py
cutwater: Thanks for pointing this out. You're right, will fix it.

cutwater:
> Style wise, does the project lean towards a preference for empty-ish init.py or with code in them? Seems to be a mix, but if pref is empty init may be worthwhile to go ahead and split these into separate py modules .

In this particular case we have quite small module and there it was converted to a package just to place configuration files within the package.

alikins:
LGTM

PR #439, opened 2018-04-09
Add 'z' mount option to bind mounts for selinux

alikins:
otherwise 'galaxy_galaxy_1' container entrypoint
script fails with perm errors like:

 "/entrypoint.sh: line 6: /galaxy/scripts/docker-dev/start-develop.sh: Permission denied"


chouseknecht:
@alikins 

Thanks for catching this!

PR #443, opened 2018-04-10
workaround 'make dev/tmux_noattach' failing in some cases

alikins:
For some cases, 'make dev/tmux' and specifically 'make
dev/tmux_noattach' will fail:

        make: *** [dev/tmux_noattach] Error 1

If 'tmux -v -v -v' is used, the error in the logfile is:

        fatal: tty_init: ttyname failed

This seems to be dependant on the version of docker
amongst other things.

https://github.com/moby/moby/issues/8755 is a similar
issue with more detail and background.

But a workaround found there
(https://github.com/moby/moby/issues/8755#issuecomment-83403289)
is to use the 'script' command to avoid the ttyname problems.

This commit uses that workaround for 'make dev/tmux'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,7 +188,7 @@ dev/tmux_noattach:
 .PHONY: dev/tmux
 dev/tmux:
 	# Connect to the galaxy container, start processes, and pipe stdout/stderr through a tmux session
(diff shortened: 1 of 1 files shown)


alikins:
I've only tested this on my fedora27 / docker 1.13.1 setup, so heads up.

alikins:
Some of CI failed, not sure why so re running. I'll merge if CI passes.

PR #446, opened 2018-04-11
Move issue_tracker_url to Repository

chouseknecht:
- Fix broken content serializers
- Add additional repo into to content serializer
- Moves issue_tracker_url from Content to Repository


galaxy/main/migrations/0090_auto_20180411_1129.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
```
cutwater: Could you please give more verbose names to migration files.

galaxy/main/migrations/0090_auto_20180411_1129.py
```diff
@@ -0,0 +1,42 @@
...
+        migrations.AddField(
+            model_name='repository',
+            name='issue_tracker_url',
+            field=models.CharField(blank=True, max_length=256,
+                                   null=True,
+                                   verbose_name=b'Issue Tracker URL'),
+        ),
+        migrations.RunSQL(
+            sql=UPDATE_REPOSITORY,
+            reverse_sql=migrations.RunSQL.noop
```
cutwater: Is this migration reversible? If no, please don't use `noop` and omit passing `reverse_sql` parameter. In this case Django will raise exception on downgrading to earlier migration.

galaxy/main/migrations/0090_auto_20180411_1129.py
chouseknecht: Done

galaxy/main/migrations/0090_auto_20180411_1129.py
chouseknecht: Done

PR #447, opened 2018-04-12
Content Detail

chouseknecht:
- Adds a content detail page.
- Adds `:namespace/:repository` route
- Adds `:namespace/:repository/:content_name` route


galaxy/api/serializers/content.py
```diff
@@ -62,7 +62,13 @@ class Meta:
 class _RepositorySerializer(serializers.ModelSerializer):
     class Meta:
         model = models.Repository
-        fields = ('id', 'name')
+        fields = (
+            'id',
+            'name',
+            'original_name',
```
alikins: 'original_name' here is referring to the upstream name of the repo?

galaxy/worker/tasks.py
```diff
@@ -102,7 +102,13 @@ def _import_repository(import_task, logger):
             logger, content_info.content_type, content_info.name)
         importer_cls = importers.get_importer(content_info.content_type)
         importer = importer_cls(context, content_info, logger=content_logger)
-
+        issue_tracker_url = ''
+        if hasattr(content_info, 'role_meta') and getattr(content_info, 'role_meta') and \
+           content_info.role_meta.get('issue_tracker_url'):
+            issue_tracker_url = content_info.role_meta['issue_tracker_url']
```
alikins: Would it make sense to make this logic a @property for galaxy.models.Repository ?

PR #458, opened 2018-04-13
Fix the APB loader

chouseknecht:
Seems changes were made to other loaders, but APB had not been updated yet. 
- Set linters
- Set content_types 
- Return a Content model, rather than dict

### DIFF
--- a/galaxy/importer/loaders/apb.py
+++ b/galaxy/importer/loaders/apb.py
@@ -21,16 +21,18 @@ import yaml
 
 from galaxy import constants
 from galaxy.importer import exceptions as exc
+from galaxy.importer import linters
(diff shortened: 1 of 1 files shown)


galaxy/importer/loaders/apb.py
```diff
@@ -20,17 +20,17 @@
 import yaml
 
 from galaxy import constants
-from galaxy.importer import exceptions as exc
+from galaxy.importer import exceptions as exc, linters, models
```
cutwater: Please  do not import more than one module per line

chouseknecht: Not disagreeing. But curious, why is it bad to import more than one per line?

PR #466, opened 2018-04-13
Validate content name on import

cutwater:


galaxy/worker/importers/base.py
```diff
@@ -57,12 +58,18 @@ def make_content(self):
             raise exc.TaskError("Content is not a repository global "
                                 "and doesn't have name.")
```
chouseknecht: This isn't something changed by this PR. However, it is something we should fix, as we talked about on the call earlier. If we're going to throw an error here, let's make sure the message is one a user will understand.

PR #470, opened 2018-04-18
API fixes for content-detail UI

chouseknecht:
- Import Tags on APBs 
- Fixes role dependencies not importing
- Fixes role platforms not importing
- Surface an external URL for repository
- Surface dependency namespace and name separately in content details
- Fixes unicode method on Content model
- Adds metadata validation for APBs based on [the APB spec doc](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/master/docs/developers.md#apb-spec-file)
- Adds unit tests for APB loader


galaxy/api/serializers/repository.py
```diff
@@ -16,6 +16,7 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 from django.core.urlresolvers import reverse
+from rest_framework import serializers as drf_serializers
```
alikins: what does 'drf' mean here?

galaxy/importer/loaders/role.py
```diff
@@ -52,8 +52,9 @@ class RoleMetaParser(object):
         'youtube': 'https://www.youtube.com/embed/{0}',
     }
 
-    def __init__(self, metadata, logger=None):
+    def __init__(self, metadata, dependencies, logger=None):
         self.metadata = metadata
+        self.dependencies = dependencies
```
alikins: If self.dependencies needs to always be list-like, I would use a optional arg and set it to empty list if None

something like

``` python
def __init__(self, metadata, dependencies=None, logger=None):
    self.metadata = metadata
    # or possible a if self.dependencies is None if it could potentially be Truthy (True, 1, 'nonempty-string", etc)
    self.dependencies = dependencies or []
 ```

File: galaxy/api/serializers/repository.py (hunk 1 above)
Comment (chouseknecht) [reply galaxy/api/serializers/repository.py]: drf = Django Rest Framework

File: galaxy/importer/loaders/role.py (hunk 2 above)
Comment (chouseknecht) [reply galaxy/importer/loaders/role.py]: Good idea. Fixed.

File: galaxy/importer/loaders/apb.py [hunk 3]
```diff
@@ -39,15 +63,27 @@ def load(self):
...
+
         try:
             name = metadata['name']
         except KeyError:
             raise exc.ContentLoadError('Missing "name" field in metadata')
+
+        try:
+            description = metadata['description']
+        except KeyError:
+            raise exc.ContentLoadError('Missing "description" field in metadata')
```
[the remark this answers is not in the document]
cutwater: > Also should probably specify the error was while importing APB metadata.

> I would also be tempted to add ContentLoaderError subclasses for each of the content types 
While I'm for adding context to importer exceptions, I don't think creating a subclass for each content type can be any useful. However it might be reasonable to create for example subclass like `MetadataParsingError`.

galaxy/importer/loaders/role.py
```diff
@@ -274,7 +274,17 @@ def _load_metadata(self):
         else:
             self.log.warning("Missing 'galaxy_info' field in metadata.")
             galaxy_info = {}
-        return galaxy_info
+
+        if 'dependencies' in metadata:
```
cutwater: Probably it's better to move all this logic to RoleMetaParser to follow single responsibility principle.

galaxy/worker/importers/apb.py
```diff
@@ -17,9 +17,36 @@
 
 from . import base
 
+from galaxy.main import models
+
 
 class APBImporter(base.ContentImporter):
+    MAX_TAGS_COUNT = 20
```
cutwater: IIRC `MAX_TAGS_COUNT` is used in role parser. Since it's more than one place we use it, please move it to constants.

galaxy/importer/loaders/apb.py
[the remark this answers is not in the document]
chouseknecht: Subclassed ContentLoaderError as suggested. Not touch other loaders, as they are outside the scope of this PR.

Not adding the metadata file name, as it is surfaced in the output log at the start of the load process, so in theory it should be apparent to the user which file is in question when an exception is thrown.

galaxy/importer/loaders/role.py
chouseknecht: Agreed. Moved.

galaxy/worker/importers/apb.py
chouseknecht: Done.

PR #473, opened 2018-04-19
Return repository type for repository detail endpoint

cutwater:
Added field `summary_fields.repository_type` for
view `galaxy.api.views.repository.RepositoryDetail`.

Repository type is a content type of stored content if
repository consists of single content item (i.e. content objects
count equals to 1).

Note: This value doesn't reflect structure of the repository.
For example both for so called "repository global role" and
repository that has `roles/` directory with a single role
repository_type will equal to `role`.

### DIFF
--- a/galaxy/api/serializers/repository.py
+++ b/galaxy/api/serializers/repository.py
@@ -55,12 +55,16 @@ class RepositorySerializer(BaseSerializer):
         if not isinstance(instance, Repository):
             return {}
         related = {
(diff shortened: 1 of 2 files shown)


chouseknecht:
LGTM.

Agree that this is 'best guess' approach. As you point out, a repo with a single content type of 'role' may be contained within a `roles` directory, and therefore not match the tradition role repo structure, but I think that's OK. I suspect we won't see many of those.

Issue #378, opened 2018-03-16
Django admin: Missing .svg files


chouseknecht:
- [x] .svg files not loading on the Admin site. 
- [x] Seeing 304 redirects on static content when loading Admin site. 

Could be a browser issue. Seeing the problem in Safari and Firefox, both after clearing cache. Haven't tried in Chrome.

cutwater:
![](https://github.trello.services/images/mini-trello-icon.png) [Django admin: Missing .svg files](https://trello.com/c/NaEBj7WM/40-django-admin-missing-svg-files)

PR #474, opened 2018-04-20
Fix django serving invalid content type for SVG

cutwater:
Issue: #378


chouseknecht:
Nice! Had not considered the mime types.

PR #480, opened 2018-04-25
Match Ansible flake8 config

chouseknecht:
Testing [atestuseraccount/galaxy-test-role](https://github.com/atestuseraccount/galaxy-test-role), which includes many database modules copied from ansible/ansible. During the import process flake8 flags the line length and `E402`.

According to the [ansible/ansible/tox.ini](https://github.com/ansible/ansible/blob/devel/tox.ini), the allowed line length is `160`, and `E402` is ignored.

This PR sets Galaxy's allowed line length to `160`, and adds `E402` to the ignore list.

### DIFF
--- a/galaxy/importer/linters/__init__.py
+++ b/galaxy/importer/linters/__init__.py
@@ -25,7 +25,7 @@ import six
 logger = logging.getLogger(__name__)
 
 LINTERS_DIR = os.path.abspath(os.path.dirname(__file__))
-FLAKE8_MAX_LINE_LENGTH = 160
+FLAKE8_MAX_LINE_LENGTH = 120
 FLAKE8_IGNORE_ERRORS = 'E402'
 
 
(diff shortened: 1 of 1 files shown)


gundalow:
for ansible/ansible we picked 160 as it avoids crazy long times and we could make our existing code pass without too much work.
If the existing Galaxy code passes 80/120/something characters then you could lower the limit

chouseknecht:
I don't have a strong opinion. I copied a bunch of existing Ansible modules into a test repo, and then imported the repo. Our static analysis kicked our a tone of `E402`s, which led to two thoughts:

1.  I don't want to try to fix them. There's probably a tool or IDE that'll do it for me, but didn't want to spend time investigating.
2. How will existing Ansible content creators react, if our line length is different than what they're used to? People get opinionated about line length. 

I was mostly concerned about #2.

chouseknecht:
Moving this to 120.

PR #500, opened 2018-05-01
Adds missing related URLs

chouseknecht:


galaxy/api/views/namespace.py
```diff
@@ -29,13 +29,17 @@
 from rest_framework.response import Response
 
 from galaxy.accounts.models import CustomUser as User
-from galaxy.main.models import Namespace, Provider, ProviderNamespace
-from .base_views import ListCreateAPIView, RetrieveUpdateDestroyAPIView
-from ..serializers import NamespaceSerializer
+from galaxy.main import models
+from . import base_views
+from .. import serializers
```
cutwater: Related imports are not recommended, but acceptable in some cases. However if you're using them please avoid imports from parent modules:

OK:
* ` from . import base_views`
* ` from .foo.bar import baz`

Not ok:
* `from .. import serializers`

chouseknecht:
@cutwater 

Thanks for pointing that out. Fixed.

Issue #489, opened 2018-04-26
My Imports long polling 


chouseknecht:
My Imports long polls to update the status of imports. It currently fetches all imports for the user, and updates the list of imports on the left side of the page. This update causes some bounce on the page.

It doesn't actually need to update all imports. Instead, it should only poll on the current import, and update the log information. All we care about is that the log for a running import gets updated in near real-time.

cutwater:
@chouseknecht is it a critical bug? It seems that it's rather a feature and not critical, since it doesn't blocks our release.

chouseknecht:
It's not critical for this release, but I feel a sense of responsibility, I suppose. Fixed by #502.

PR #502, opened 2018-05-02
Fixes for my-imports

chouseknecht:
- Removes continual page refresh. Now only refresh the selected item, when the status in pending or running.
- Adds 'scroll to end of log' widget
- Adds tool tips
- Sets cursor style to pointer, when hovering over list item
- Establishes enums subdir


chouseknecht:
Fixes #489

PR #504, opened 2018-05-02
Fix psycopg2 warning message

cutwater:
This patch fixes warning raised by psycopg2 lib:

> The psycopg2 wheel package will be renamed from release 2.8;
> in order to keep installing from binary please use
> "pip install psycopg2-binary" instead.
> For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.


chouseknecht:
\o/

Issue #497, opened 2018-04-30
Don't import nested modules/plugins


chouseknecht:
In our current atestuseraccount/ansible-testing-content repo, we currently have a nested structure under `library`. 

- [x] Remove the nested structure, because it's not really supported by Ansible at runtime. 
- [x] Remove the ability of the import process to recurse subdirectories under `library`, and the `*_plugin` directories
- [x] During import, show a warning, if a nested structure is detected in the module or plugin path

PR #509, opened 2018-05-03
Flat directory structure for modules and module utils.

cutwater:
Module finder now supports only flat structure of python modules.
Finder writes a warning message if a nested directory detected in
modules directory.

Issue: #497

### DIFF
--- a/galaxy/importer/finders.py
+++ b/galaxy/importer/finders.py
@@ -16,7 +16,6 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 import collections
-import glob
 import itertools
 import logging
 import os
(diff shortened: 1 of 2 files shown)


galaxy/importer/finders.py
```diff
@@ -102,13 +101,18 @@ def _find_content(self):
...
-                rel_path = os.path.relpath(file_path, self.path)
-                yield Result(content_type, rel_path, extra={})
+        for file_name in os.listdir(content_dir):
+            file_path = os.path.join(content_dir, file_name)
+            if os.path.isdir(file_path):
+                self.log.warning("Directory detected: '{0}'. "
+                                 "Nested modules are not supported.")
+                continue
+            if (not os.path.isfile(file_path)
+                    or not file_name.endswith('.py')
```
chouseknecht: Are we only importing files that end with `.py`?  Ansible modules don't have to be Python.

chouseknecht: Because Ansible modules are not required to be Python, we need (at some point) to add a MIME type (or some kind of indication), if possible, that we can use to tell the user what language the module is written in.

bcoca:
module_utils does support 'nested' directories, this is valid:

`from ansible.module_utils.maindir.subdir import X`

galaxy/importer/finders.py
cutwater: Right, but first we should determine which file types we gonna support.

cutwater:
> module_utils does support 'nested' directories, this is valid:
We don't support any nested content items at least right now.
As possible solution we can import modules and packages as a flat items. So

`ansible.module_utils.maindir.subdir`

will be imported as module_utils content named `maindir` containing python package and all nested modules.
For 3.0 we're going to support only flat modules and packages support will be added in the future release.

bcoca:
module_utils is for modules to consume, not to contain them, modules go in `library/`

chouseknecht:
Merging. We'll revisit nested module_utils next release.

PR #513, opened 2018-05-08
Set plugin description

chouseknecht:


galaxy/importer/loaders/plugin.py
```diff
@@ -56,8 +56,13 @@ def make_name(self):
     def load(self):
         self._parse_plugin()
 
+        description = ''
+        if self.documentation:
+            description = self.documentation.get('short_description', '')
+
         return models.Content(
             name=self.name,
+            description=description,
```
alikins: I think this LGTM but had a question.

Will the Content.description being redundant with Content->metadata->>'short_description' be problem?

Seems like it should be fine.

And since Content.description is not nullable but Content->metadata->>'short_description' is, I guess they do have different uses.

Especially if Content.description could become user editable from the web side at some point (as opposed to requiring a change to the users plugin code and reimport).

PR #518, opened 2018-05-11
Adds content_counts to Namespace

chouseknecht:
For each namespace, show a count of the related content items by type.

```
"content_counts": {
    "action_plugin": 1,
    "module_utils": 6,
    "lookup_plugin": 3,
    "module": 8,
    "role": 6,
    "filter_plugin": 2,
    "strategy_plugin": 3
},
```

### DIFF
--- a/galaxy/api/serializers/namespace.py
+++ b/galaxy/api/serializers/namespace.py
@@ -16,7 +16,7 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 from django.core.urlresolvers import reverse
-from galaxy.main.models import Namespace
+from galaxy.main import models
 from . import serializers
 
 
@@ -26,9 +26,8 @@ __all__ = [
 
 
 class NamespaceSerializer(serializers.BaseSerializer):
-
     class Meta:
-        model = Namespace
+        model = models.Namespace
         fields = serializers.BASE_FIELDS + (
             'id',
             'description',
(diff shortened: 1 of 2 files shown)


chouseknecht:
@cutwater 

Figured out how to do a `group_by` from the Namespace model object. It now does a single query for each namespace, rather than multiple queries per Namespace (i.e., one query for each possible content type).

galaxy/api/serializers/namespace.py
```diff
@@ -58,9 +58,15 @@ def get_summary_fields(self, instance):
             'provider': pn.provider.id,
             'provider_name': pn.provider.name.lower()
         } for pn in instance.provider_namespaces.all()]
+
+        content_counts = {}
```
cutwater: You can use dict expression here:
```
content_counts = {c['content_type__name']: c['count'] for c in instance.content_counts()}
```

galaxy/api/views/namespace.py
```diff
@@ -39,7 +39,7 @@
     'NamespaceProviderNamespacesList',
     'NamespaceProviderNamespacesList',
     'NamespaceContentList',
-    'NamespaceOwnersList',
```
cutwater: Why did you remove trailing comma?

galaxy/main/models.py
```diff
@@ -15,9 +15,12 @@
 # You should have received a copy of the Apache License
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
+import logging
+
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.db import models
+from django.db.models import Count
```
cutwater: Module `models` already imported on line 22.

galaxy/api/views/namespace.py
chouseknecht: Fixed

galaxy/main/models.py
chouseknecht: Fixed

PR #519, opened 2018-05-11
Fix import error 'Context' object has no attribute 'versions'

alikins:
Error seen in ui import logs:
        Error identifying tags to remove: 'Context' object has no attribute 'versions'

_update_repository_versions is passed a Context object that
has a repository, update method to use Context.repository


cutwater:
Fixed by #522

Issue #498, opened 2018-04-30
User defined role and repository names


chouseknecht:
As an author, it should be easy to set the role or repository name for my content.

Define a consistent method of determining a repository \ role name that can be used by the Galaxy CLI, and Pulp.

**Acceptance criteria:**
* The UI should not be used to set the repo name, or role name.
* User should not need to log into Galaxy, in order to trigger an import.
* Single import User shouldn't have to import something 2x, in order to get it named correctly.
* Role contributors should be able to set the role name.
* Role name should be set in metadata, and metadata should be the single source of truth for alternate name.

**Action Items:**
_TBD_

---
- [x] When importing a traditional role, where a `meta/main.yml` exists at the root of the project, the following will be used to set the role name on the Galaxy server:
    - Use `galaxy_info.role_name`, if set in `meta/main.yml`
    - ~~When `galaxy_info.role_name` is not set, then start with the SCM repository name, and remove /^ansible-role[-_+.]/, remove /^ansible[-_+.]/, and remove /^role[-_+.]/~~

- [ ] When importing a role from a multi-content repository, the following will be used to set the role name on the Galaxy server:
    - ~~Use `galaxy_info.role_name`, if set in `role_directory_name/meta/main.yml`~~
    - ~~When `galaxy_info.role_name` is not set, then start with the role's directory name , and remove /^ansible-role[-_+.]/, remove /^ansible[-_+.]/, and remove /^role[-_+.]/~~

- [ ] The UI will not attempt to override the repository or role name, ~~nor will it allow the user to override a repository or role name. The logic for setting the repository and role name will exist only in the API.~~

- [x] The UI will continue to allow the user to override the Repository name

- [ ] When the UI submits a repository for import, it will receive the repository `original_name` and `name` attributes, and surface them accordingly 

- [ ] ~~Update the role `init` templates in Galaxy CLI to include `galaxy_info.role_name`~~

chouseknecht:
@alikins @daviddavis

Let me know, if you have questions or comments regarding the above.

daviddavis:
LGTM 👍

chouseknecht:
We still have this wrong. 
- The UI should not be used to set the repo name, or role name. You should not need to log into Galaxy, in order to trigger an import, and you shouldn't have to import something 2x, in order to get it named correctly.
- Role contributors expect to be able to set the role name.
- Role name should be set in metadata, and metadata should be the single source of truth.

chouseknecht:
[...]

PR #536, opened 2018-05-15
 [DONOTMERGE] Allow role name alias defined in `meta/main.yml`

cutwater:
Issue: #498


cutwater:
This PR is not valid anymore as a result of BlueJeans discussion on May 15th.

cutwater:
RepositoryType should be renamed to RepositoryFormat, stored to database and pushed as a standalone PR.

PR #537, opened 2018-05-15
Use json serializer for celery tasks

cutwater:


chouseknecht:
This makes sense; however, it points to the fact that we really, really need to add API unit testing.

PR #541, opened 2018-05-16
Detect repository format when importing a repository and store it in DB

cutwater:


galaxy/api/serializers/repository.py
```diff
@@ -28,16 +28,15 @@
...
     external_url = drf_serializers.SerializerMethodField()
-    repository_type = drf_serializers.SerializerMethodField()
 
     class Meta:
         model = Repository
         fields = serializers.BASE_FIELDS + (
             'id',
             'original_name',
             'description',
+            'format',
```
chouseknecht: Why are we changing this to `format`? This breaks things in the UI.

chouseknecht:
After this merges, the UI needs to be updated. The UI is looking for "repository_type".

chouseknecht:
NVM. I see that @cutwater included the UI changes already. Cool!

Issue #433, opened 2018-04-09
Limit Namespace creation and deletion to Staff


chouseknecht:
There's concern that anyone can log into Galaxy and create any Namespace they like. This leaves us vulnerable to folks creating namespaces for which they have no actual copyright or claim.

- [x] Disable the Add button on My Content when the user is not a staff member
- [x] Namespace owners can enable/disable a namespace, but they cannot delete. Add a delete button on My Content that is only enabled when the user is a staff member.
- [x] Add a route guard for /my-content/add. Only staff members should be allowed to navigate to the page.

chouseknecht:
Also, for admins, don't require provider namespaces to be associated with the Namespace. Will need to test what happens with Add Content when no provider namespaces are available.

duckie:
Hi guys

That's well and great, but how can I create a namespace for my company ? Do you have a dedicated process for it, an email address I can write to ?

PR #574, opened 2018-05-22
Add is_vendor attribute to Namespace

### LINKED ISSUE #555: As a system admin, I want to indicate that a Namespace represents a partner, so that we can highlight vendor-produced content
In the UI, we want to differentiate community contributed content from vendor contributed content. To support this, we need to know which Namespaces represent a partner. 

- [ ] Add 'partner' boolean attribute to Namespace
- [ ] Surface 'partner' in the Namspace serializer. Check Content, Repository, Provider_Namespaces, and surface 'partner' where Namespace is part of Summary Fields
- [ ] Default to False for all existing Namespaces

alikins:
Adds Namespace 'vendor' boolean #555

Default to False.

Include the field in NamespaceSerializer fields.

Add the 'partner' field to summary_fields['namespace']
for Repository, ProviderNamespace. Content
uses local _NamespaceSerializer so add the field to
cls.fields there.

Add 0098_namespace_partner.py migration

Fixes #555


tima:
Perhaps we should use the term "vendor" (any company or organization even) here so we have the option of using "partner" (there is a more formal relationship established between Red Hat Ansible and the vendor) later.

galaxy/api/views/namespace.py
```diff
@@ -258,7 +258,7 @@ def update(self, request, *args, **kwargs):
                 owners.append(request.user.pk)
             update_owners(instance, owners)
 
-        for item in ('name', 'description', 'avatar_url', 'location', 'company', 'email', 'html_url', 'active'):
+        for item in ('name', 'description', 'avatar_url', 'location', 'company', 'email', 'html_url', 'active', 'partner'):
```
chouseknecht: I think 'partner' needs to be added above, on line 215, in the 'post' method of NamespaceList. It wouldn't let me add a comment to that line directly, but take a look.

chouseknecht:
From a code and functionality standpoint this looks good to me. One small comment in the namespace view.

Sent @tima a follow-up email asking for resolution on 'partner' vs 'vendor'. Good question, and something we should definitely resolve now, if possible.

chouseknecht:
@alikins 

From @tima, please change `partner` to `vendor`.

galaxy/api/views/namespace.py
alikins: Added.

alikins:
rebased and amended to squash migrations to single 0098_namespace_vendor.py script.

galaxy/main/models.py
```diff
@@ -540,6 +540,8 @@ class Meta:
         verbose_name="Web Site URL"
     )
 
+    vendor = models.BooleanField(default=False)
```
cutwater: Can you please name it `is_vendor`

galaxy/api/views/namespace.py
```diff
@@ -258,7 +258,7 @@ def update(self, request, *args, **kwargs):
                 owners.append(request.user.pk)
             update_owners(instance, owners)
 
-        for item in ('name', 'description', 'avatar_url', 'location', 'company', 'email', 'html_url', 'active'):
+        for item in ('name', 'description', 'avatar_url', 'location', 'company', 'email', 'html_url', 'active', 'vendor'):
```
cutwater: Probably we shouldn't update `vendor` field from API. It is only to be managed by admin users.

chouseknecht: By 'admin' we mean user.is_staff=True, not the Django admin. I think what @alikins has is correct. 

There is a separate issue, #433, that covers limiting Namespace create/delete to users where user.is_staff. We'll add the permission checks as part of that issue.

cutwater: Thanks for the clarification.

chouseknecht:
Changed 'vendor' to 'is_vendor'.

PR #579, opened 2018-05-24
Improve Python 3 compatibility

kbugaienko:


galaxy/api/filters.py
```diff
@@ -93,10 +95,10 @@ def to_python_boolean(self, value, allow_none=False):
             return None
         else:
             raise ValueError(
-                u'Unable to convert "{}" to boolean'.format(unicode(value)))
+                u'Unable to convert "{}" to boolean'.format(six.text_type(value)))
```
cutwater: You can safely drop explicit casting to unicode here. Since unicode literal is used, value will be implicitly converted to unicode.

galaxy/api/serializers/serializers.py
```diff
@@ -102,10 +103,10 @@ def get_fields(self):
         for key, field in ret.items():
             if key == 'id' and not getattr(field, 'help_text', None):
                 field.help_text = u'Database ID for this {}.'.format(
-                    unicode(opts.verbose_name))
+                    six.text_type(opts.verbose_name))
```
cutwater: ditto

galaxy/api/serializers/serializers.py
```diff
@@ -102,10 +103,10 @@ def get_fields(self):
...
             elif key == 'url':
                 field.help_text = u'URL for this {}.'.format(
-                    unicode(opts.verbose_name))
+                    six.text_type(opts.verbose_name))
```
cutwater: ditto

galaxy/api/serializers/serializers.py
```diff
@@ -119,12 +120,12 @@ def get_fields(self):
             elif key == 'created':
                 field.help_text = (
                     u'Timestamp when this {} was created.'.format(
-                        unicode(opts.verbose_name)))
+                        six.text_type(opts.verbose_name)))
```
cutwater: ditto

galaxy/api/serializers/serializers.py
```diff
@@ -119,12 +120,12 @@ def get_fields(self):
...
                 field.type_label = 'datetime'
             elif key == 'modified':
                 field.help_text = (
                     u'Timestamp when this {} was last modified.'.format(
-                        unicode(opts.verbose_name)))
+                        six.text_type(opts.verbose_name)))
```
cutwater: ditto

galaxy/main/celerytasks/tasks.py
```diff
@@ -127,7 +128,7 @@ def refresh_user_repos(user, token):
         user.cache_refreshed = True
         user.save()
         msg = u"User {} Repo Cache Refresh Error: {}".format(
-            user.username, unicode(exc))
+            user.username, six.text_type(exc))
```
cutwater: ditto

galaxy/main/celerytasks/tasks.py
```diff
@@ -137,7 +138,7 @@ def refresh_user_repos(user, token):
...
```
cutwater: ditto

galaxy/main/celerytasks/tasks.py
```diff
@@ -147,7 +148,7 @@ def refresh_user_repos(user, token):
         user.cache_refreshed = True
         user.save()
         msg = u"User {} Repo Cache Refresh Error: {}".foramt(
-            user.username, unicode(exc))
+            user.username, six.text_type(exc))
```
cutwater: ditto

galaxy/main/celerytasks/tasks.py
```diff
@@ -168,21 +169,21 @@ def refresh_user_stars(user, token):
     try:
         gh_api = github.Github(token)
     except github.GithubException as exc:
-        msg = u"User {} Refresh Stars: {}".format(user.username, unicode(exc))
+        msg = u"User {} Refresh Stars: {}".format(user.username, six.text_type(exc))
```
cutwater: ditto

galaxy/main/celerytasks/tasks.py
```diff
@@ -168,21 +169,21 @@ def refresh_user_stars(user, token):
...
         LOG.error(msg)
         raise Exception(msg)
 
     try:
         ghu = gh_api.get_user()
     except github.GithubException as exc:
-        msg = u"User {} Refresh Stars: {}".format(user.username, unicode(exc))
+        msg = u"User {} Refresh Stars: {}".format(user.username, six.text_type(exc))
```
cutwater: ditto

galaxy/main/celerytasks/tasks.py
```diff
@@ -168,21 +169,21 @@ def refresh_user_stars(user, token):
...
         LOG.error(msg)
         raise Exception(msg)
 
     try:
         subscriptions = ghu.get_subscriptions()
     except github.GithubException as exc:
-        msg = u"User {} Refresh Stars: {}".format(user.username, unicode(exc))
+        msg = u"User {} Refresh Stars: {}".format(user.username, six.text_type(exc))
```
cutwater: ditto

galaxy/main/celerytasks/tasks.py
```diff
@@ -205,7 +206,7 @@ def refresh_user_stars(user, token):
     try:
         starred = ghu.get_starred()
     except github.GithubException as exc:
-        msg = u"User {0} Refresh Stars: {1}".format(user.username, unicode(exc))
+        msg = u"User {0} Refresh Stars: {1}".format(user.username, six.text_type(exc))
```
cutwater: ditto

galaxy/main/celerytasks/tasks.py
```diff
@@ -268,7 +269,7 @@ def refresh_role_counts(start, end, token, tracker):
             else:
                 passed += 1
         except Exception as exc:
-            LOG.error(u"FAILED: {0} - {1}".format(full_name, unicode(exc)))
+            LOG.error(u"FAILED: {0} - {1}".format(full_name, six.text_type(exc)))
```
cutwater: ditto

galaxy/main/celerytasks/tasks.py
```diff
@@ -307,5 +308,5 @@ def clear_stuck_imports():
             ri.save()
             transaction.commit()
     except Exception as exc:
-        LOG.error(u"Clear Stuck Imports ERROR: {}".format(unicode(exc)))
+        LOG.error(u"Clear Stuck Imports ERROR: {}".format(six.text_type(exc)))
```
cutwater: ditto

galaxy/main/models.py
```diff
@@ -56,7 +57,7 @@ class Meta:
 
     def __unicode__(self):
         if hasattr(self, 'name'):
-            return unicode("%s-%s" % (self.name, self.id))
+            return six.text_type("%s-%s" % (self.name, self.id))
```
cutwater: It would be better to rewrite it as

```
u'{}-{}'.format(self.name, self.id)
```

galaxy/main/models.py
```diff
@@ -184,7 +185,7 @@ class Meta:
     )
 
     def __unicode__(self):
-        return unicode("%s (alias of %s)" % (
+        return six.text_type("%s (alias of %s)" % (
```
cutwater: Same here. Please use unicode string literal and new style formatting.

galaxy/common/version.py
```diff
@@ -14,7 +14,7 @@
 #
 # You should have received a copy of the Apache License
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
-
+from __future__ import unicode_literals
```
webknjaz: This specific import is considered harmful, thus we don't use it in Core. 

however, you can use this snippet for py3fying some other bits:
```python
from __future__ import absolute_import, division, print_function
__metaclass__ = type
```

galaxy/main/migrations/0083_content_search_vector.py
```diff
@@ -52,7 +52,7 @@ class Migration(migrations.Migration):
         migrations.AddIndex(
             model_name='content',
             index=psql_indexes.GinIndex(
-                fields=[b'search_vector'],
+                fields=['search_vector'],
```
webknjaz: this is probably a breaking change

galaxy/common/version.py
cutwater: I strongly disagree on using `__metaclass__ = type` magic. It implicitly changes behavior of python classes, while explicit inheritance from `object` class works both for Python 2 and 3.

webknjaz: FWIW it makes every class in the module new-style. We even have a linter for ansible modules to require these two lines.

galaxy/main/migrations/0083_content_search_vector.py
cutwater: Apparently it's not. Checked both on 2 and 3.

galaxy/common/version.py
cutwater: Yep, but I am not aware of any other big python project that uses this approach. Unlike `__future__` imports that are standardized, this compatibility import looks more like a hack, since it is not widely used (project specific), implicit, global (on module level) and not required at all when using explicit inheritance from `object` class.
Considering all the above in my opinion it is a bad practice, which we are unlikely to follow.

PR #587, opened 2018-05-28
Python 3: Fix __unicode__ methods and use new-style string formatting

kbugaienko:


chouseknecht:
@kbugaienko 

Thanks for taking the time to work on this. Nice job!

Issue #556, opened 2018-05-19
Verify and improve content download count calculation


chouseknecht:
As a user, I expect Download count to be reasonably accurate, as it impacts search result relevance

- [x] Track download count at the repository level. It's currently a Content attribute
- [x] Coordinate between old client, and new client how to determine a download.
- [ ] ~~Review existing throttle code. Research, should we keep it, can we make it better?~~

alikins:
For throttling, one approach that could help is having the client send some per system info along with http requests (for ex, generating a uuid on first use of client on a system and sending it along with each request in the future).

That should help distinguishing between multiple systems that end up with the same ip address.  In production, it will also help troubleshooting any issues that seem to effect particular clients (assuming this info also makes it in the logs)

cutwater:
> For throttling, one approach that could help is having the client send some per system info along with http requests (for ex, generating a uuid on first use of client on a system and sending it along with each request in the future).

This approach can give more accurate results than throttling by IP addresses, however I am worried about stateless CI systems. In case somebody executes client from container, UUID will be different on each run.

P.S. Maybe it's worth to think in different direction. PyPi removed download counts at all and they described their reasons behind this decision: https://mail.python.org/pipermail/distutils-sig/2013-May/020855.html. Some of them seem to me legitimate in context of galaxy.

chouseknecht:
@cutwater 

Based on those examples, my thinking is that we do 2 things:

1. Remove any attempt at throttling.
2. Track it at the repo level
3. Track over last 30 days. Maybe something was popular and/or maintained in the past, but now it's not
4. Make it easy to get total, overall downloads for the last 30 days.

Before we do any work on this, let's review as a team. Adding as a topic to the weekly server meeting agenda.

alikins:
@cutwater Looking at this again, I think my comments at https://github.com/ansible/galaxy/issues/556#issuecomment-390673783 doesn't make sense for this download_count (the one potentially displayed to users). The per system uuid case would be better suited for abuse detection, but doesn't seem necessary for user facing info.

chouseknecht:
@newswangerd 

Review the UI, and make sure we're surfacing download count correctly. It should show up on Search and Content Details. On Content Details it may actually be hardcoded to 0, as we were waiting for it to land on the Repository object and /api/v1/repositories endpoints.

chouseknecht:
@newswangerd 

[...]

PR #593, opened 2018-05-29
Move content download count to repository level

cutwater:
Issue: #556


chouseknecht:
Waiting on merging this until we decide what to do with the following:

- throttling
- tracking download over a period of time, i.e. 'last 30 days'

PR #597, opened 2018-05-29
Python 3: Fix new-style string formatting

kbugaienko:


chouseknecht:
@kbugaienko, thanks for submitting this. Nice job!

@cutwater, in ansible/ansible, one of the sanity checks they run is testing whether or not format strings include numbers between the braces. For example, they flag `"{}".format(foo)` as an error, and pass `{0}.format(foo)`. Is this a standard, and should we enforce it as well?

cutwater:
@chouseknecht Specifying position numbers in braces was required before python 2.7. Since ansible/ansible has to support python 2.6, it makes sense for them to enforce that. Galaxy supports only python 2.7, so we are free to use both styles. I'm not aware of any formal recommendations on this matter.
Personally I prefer using `{}` style unless it's required to have non-sequential order or include value in format string multiple times.

PR #601, opened 2018-05-30
Search name using partial match and OR

chouseknecht:
Construct Namespace filter using `__icontains` and `OR`

### DIFF
--- a/galaxy/api/views/search.py
+++ b/galaxy/api/views/search.py
@@ -19,7 +19,9 @@ import operator
 from collections import OrderedDict
 
 import six
-from django.db.models import F, Func, Value, Count, ExpressionWrapper
+import operator
+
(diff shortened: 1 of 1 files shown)


galaxy/api/views/search.py
```diff
@@ -154,7 +154,11 @@ def add_tags_filter(queryset, tags):
     def add_namespaces_filter(queryset, namespaces):
         if not namespaces:
             return queryset
-        return queryset.filter(namespace__name__in=namespaces)
+        queries = [Q(namespace__name__icontains=name) for name in namespaces]
+        query = queries.pop()
```
cutwater:
```
import six
import operator
# ...
query = six.reduce(operator.or_, queries)
```

PR #602, opened 2018-05-30
Full access for staff members

chouseknecht:
API changes to allow user's with `is_staff` full access to create/update/delete Namespace, Provider Namespace and Repository objects.

Plus:

- Add description to content_objects returned by Repository serializer 

Issue #433


galaxy/api/views/provider_source.py
```diff
@@ -54,6 +54,7 @@ def get(self, request, *args, **kwargs):
                         'id': provider.pk,
                         'name': provider.name.lower()
                     }
+                    print(source['name'])
```
cutwater: Please remove debug print.

PR #603, opened 2018-05-30
UI changes for allowing staff full access to Namespace

chouseknecht:
UI changes to allow user's with `is_staff` full access to create/update/delete Namespace, Provider Namespace and Repository objects.

- Upgraded to latest patternfly and latest Ng CLI tool.
- On Search page, added sort by fork count
- Added fork count to Search results 
- Created a Copy-n-Paste component, and applied it anywhere an `install` command appears.
- Fixed at least one thing causing a performance issue on My Content

Issue #433 
Requires #602


galaxy/api/views/provider_namespace.py
```diff
@@ -42,6 +42,9 @@
 
 
 def check_provider_access(provider, user, name):
+    if user.is_staff:
+        return True
+
```
alikins: check_provider_access() checks this itself but check_namespace_access() doesnt.
Any particular reason?

Nothing seems to use the return of this method, does it need to 'return True' or could it just 'return' ?

chouseknecht:
@alikins 

Thanks for pointing out my hackery. Turns out there isn't any need to check `is_staff` on Provider Namespace. In the UI, the user modifies Namespace and Repositories, and Provider Namespace gets created and modified indirectly by the server. Reverted the change.

PR #605, opened 2018-05-30
Add download_url field to repository and version api endpoints

cutwater:


alikins:
Patch looks good to me.

For a question not specific to this pr but inspired by it...
Will we eventually ever also need download_url to be per Content? 

I could see future cases where we could need to get the url for the scm repo archive [Repository.get_download_url()] and maybe also a download url for each bit of content
in it. I don't think we need per content download url for now though and when/if
we do that will likely be a part of Pulp integration so nothing urgent.

cutwater:
@alikins While I agree that per content download url could be useful in some cases, I hardly imagine how it will look like from implementation perspective. Currently we rely on GitHub \ GItLab archive feature, so we can simply fetch that URL to retrieve a repository. 
However when we're speaking about content, first thing we should keep in mind that contents are not only plain files. So even if we can generate such URL for plain file on GitHub \ GitLab which can be used for fetching files, I don't see a simple way how to fetch a directory without cloning an entire repository.
I'd like to move this discussion out of PR comments, so we don't lose it. Can you please create a proposal issue.

PR #633, opened 2018-06-06
Stop installing packages 2x

chouseknecht:


Makefile
```diff
@@ -57,15 +57,17 @@ clean:
...
-	cd galaxyui; ng lint
-	rm -rf galaxyui/node_modules
 
 .PHONY: build/dist
 build/dist: build/static
 	python setup.py clean bdist_wheel
 	GALAXY_VERSION=$$(python setup.py --version) \
 		&& ln -sf galaxy-$$GALAXY_VERSION-py2-none-any.whl dist/galaxy.whl
 
+.PHONY: build/jslint
```
cutwater: We use `build` prefix for jobs that do build. Please use `test` prefix.

PR #667, opened 2018-06-14
Re implement download count

### LINKED ISSUE #556: Verify and improve content download count calculation

newswangerd:
- Fix views to display download count from the repository database
- Increment download count when request is made to the repository API

fixes #556

### DIFF
--- a/galaxy/api/views/roles.py
+++ b/galaxy/api/views/roles.py
@@ -20,7 +20,7 @@ import logging
 from rest_framework.response import Response
 from django.http import Http404
 
-from galaxy.main.models import Content
+from galaxy.main.models import Content, Repository
 
 from .views import filter_role_queryset
 from .base_views import ListAPIView, RetrieveAPIView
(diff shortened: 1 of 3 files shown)


galaxy/api/views/roles.py
```diff
@@ -53,6 +53,12 @@ def list(self, request, *args, **kwargs):
             qs = self.get_queryset()
             qs = qs.filter(**params)
             page = self.paginate_queryset(qs)
+
+            if request.query_params.get('name') and qs.count() == 1:
```
cutwater:
```
 if request.query_params.get('name'):
    content = qs.first()
    if content is not None:
    ...
```

Issue #692, opened 2018-06-21
Disallow additional special chars in Namespace and Repository names


chouseknecht:
From the content namespace discussion with the core team, Galaxy server must enforce the following:

For namespace names:
- Only accept word characters. No '-' or '.' are allowed.
- Enforce a minimum length of 2 characters
- Don't allow leading '_'

For repository names:
- Convert '-' and '.' to '_', so that the resulting repository name contains word chars only
- Enforce a minimum length of 2 characters
- Don't allow leading '_'

Test from the UI side that the above is true, and that useful errors are surfaced.

newswangerd:
@chouseknecht  if we disallow '-' and '.' in namespace names, won't that make it impossible for some people to create accounts on galaxy?

chouseknecht:
See comment on #429 for repository/role import fixes 

Fix account creation during login. This is the login signal handler. Replace the if statement on line #44 with logic something like the following: 

- Get the user's social login, e.g. newswanger-test
- Convert social login to lowercase, and replace [-.] to _, giving newswanger_test
- Check to see if a namespace matching the converted social login exists. If so, exit
- Create a namespace for the converted social login

Issue #662, opened 2018-06-13
New users already in a namespace don't get their default namespace created


newswangerd:
When new users join galaxy and are already in a github organization that has a namespace on galaxy, no namespace gets created for the user.

chouseknecht:
This is getting fixed as part of #692

Issue #429, opened 2018-04-09
Verify API and import validation limits names to [\w] chars, and enforces data integrity


chouseknecht:
It should not be possible to create a namespace name, repository name, nor content name containing special characters.

- [x] Verify namespace.name only accepts [\w-], and that a validation error is surfaced by the API
- [x] ~~If the API encounters one or more '.' in the repository name, it should replace all '.' with '_', and store the new name as 'name', and the original name as 'original_name'.~~
- [x] Verify repository.name only accepts [\w-], and that a validation error is surfaced by the API
- [x] Verify that the import process throws a validation error, when content.name contains a character not in [\w-]
- [x] Verify there is a unique constraint on lowercase(namespace.name), and that a violation is surface by the API 
- [x] Verify there is a a unique constraint on lowercase(namespace.name, repository.name), and a violation is surfaced by the API
- [x] Verify the API enforces a unique constraint on lowercase(namespace.name, repository.name, content.name), and a violation is surfaced by the API

chouseknecht:
Here's the query with regex used to identify invalid names currently in production (see #299): 

```
SELECT name, modified FROM main_content WHERE name !~ '^[\w-]+$';
```

newswangerd:
last bullet fails. See #664

chouseknecht:
Closed #664. If two roles or content items in a repo collide, I think for now, that's on the user, as it's difficult for us to detect. We may address this in the future with Ansible Lint or some other tool.

chouseknecht:
To fix the last bullet item we need to make the following changes:

- [ ] in worker/tasks.py set the repository.name, to original_name lowercased and with [-.] converted to '_'.
- [ ] In worker/importers/base.py, change match expression on line #62 by removing '-'
- [ ] in main/models.py change the unique index on Content to include repository.name

Issue #39, opened 2017-10-13
Unable to refresh repositories and import role with a specific user


chouseknecht:
_From @shibayu36 on September 13, 2017 7:28_

I use galaxy's user named MACKEREL.  When I log in this user, I cannot "Refresh Your Repositories from GitHub" and "Import Role" from https://galaxy.ansible.com/roleadd#/ page.

I saw Network tab in chrome developer tools, then I found a refresh endpoint( https://galaxy.ansible.com/api/v1/repos/refresh/ ) returns Internal Server Error.

I thought this problem occurs because this account named MACKEREL have been connected to multiple Github accounts.  So I revoked Ansible Galaxy from Github Apps(https://github.com/settings/applications) once, and reconnected MACKEREL user to github, but this problem was not fixed yet.

What should I do?

_Copied from original issue: ansible/galaxy-issues#282_

chouseknecht:
@shibayu36 

I'm confused. If I look at https://github.com/mackerel, there are 0 repositories. Are the repositories part of an organization? And if so, is Galaxy authorized with the organization?

chouseknecht:
_From @shibayu36 on October 10, 2017 2:34_

@chouseknecht 

We got MACKEREL user a long time ago (maybe in 2015).  And if I remember correctly, we could decide an account name freely at that time.  So our account name is MACKEREl, but our repositories are in https://github.com/mackerelio .  (https://github.com/mackerel is other user's account whom I don't know).

Now, I think MACKEREL is connected to multiple github users in the mackerelio organization.  For example, mackerel-bot, Songmu and so on.  They are members in the mackerelio organization (https://github.com/orgs/mackerelio/people).

PR #712, opened 2018-06-26
Sanitize repo and namespace names

newswangerd:
Fixes: #429 
Fixes: #692 
Fixes: #662

Namespaces:
- Convert existing namespaces with `-` in the name to `_`
- Ensure names don't start with `_` and are more than 2 characters
- Lowercase namespaces for new github users

Repos:
- Lowercase name and replace `-` and `.` with `_` during new repo creation
- Ensure names don't start with `_` and are more than 2 characters

Content:
- For single content repositories, set content name to repo.original name
- Added repository to unique fields


galaxy/main/signals/handlers.py
```diff
@@ -41,15 +41,17 @@ def user_logged_in_handler(request, user, **kwargs):
...
 
-    if user.namespaces.count():
-        # User is associated with one or more namespaces.
-        # Even though the Namespaces may be inactive, we'll assume
-        # the user has been here before, and knows how to manage Namespaces.
+    username = social.extra_data.get('login')
+    sanitized_username = username.lower().replace('-', '_')
+
+    try:
+        namespace = models.ProviderNamespace.objects.get(name=username)
```
chouseknecht: Think the filter should be name__iexact=username, to account for existing namespaces that may have contain uppercase letters.

galaxy/main/signals/handlers.py
```diff
@@ -59,7 +61,7 @@ def user_logged_in_handler(request, user, **kwargs):
 
         if provider.name.lower() == 'github':
```
chouseknecht: I think we no longer need to iterate over Providers. We assumed GitHub above to get the social object (line #39), so might as well continue to assume GitHub here as well, and use the existing social object.

galaxy/worker/importers/base.py
```diff
@@ -50,8 +50,10 @@ def make_content(self):
         repo = self.ctx.repository
         ns = repo.provider_namespace.namespace
 
-        name = repo.name
-        original_name = repo.name
+        # Name is the content name, which in the case of multi content repos
+        # might not be the same as the repository name
+        name = repo.original_name
```
chouseknecht: Name should default to the repo.name, which should have been sanitized of any '-' and '.' chars and converted to lowercase() somewhere.

galaxy/main/migrations/0108_sanitize_namespace_names.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.13 on 2018-05-31 13:18
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+STRIP_VERSION_SUFFIX = """
```
chouseknecht: Nit picky, I know, but this variable name should be related to the task at hand. Look like it was copy-pasta'd from some other migration.

galaxy/api/views/namespace.py
```diff
@@ -48,9 +48,13 @@
 def check_basic(data, errors):
     if not data.get('name'):
         errors['name'] = "Attribute 'name' is required"
-    elif not re.match('^[\w-]+$', data['name']):
-        # Allow only names containing word chars and '-'
-        errors['name'] = "Name can only contain [A-Za-z0-9-_]"
+    elif not re.match('^[\w]+$', data.get('name')):
+        # Allow only names containing word chars
+        errors['name'] = "Name can only contain [A-Za-z0-9_]"
+    elif(len(data.get('name')) <= 2):
```
cutwater: Possible `TypeError` here. You're using `data.get('name')` multiple times.
Use `name = data.get('name')` in the beginning of the function and work with a variable below.

galaxy/api/views/namespace.py
```diff
@@ -48,9 +48,13 @@
...
+        errors['name'] = "Name must be longer than 2 characters"
+    elif(data.get('name')[0] == '_'):
```
cutwater: `name.startswith()`

galaxy/api/views/repository.py
```diff
@@ -62,13 +62,17 @@ def get_repo(provider_namespace, user, repo_name):
 def check_name(name):
     if not name:
         raise ValidationError(detail={'name': 'Name is required'})
+    elif not re.match('^[\.\w-]+$', name):
```
cutwater: No need for `elif`, raise will interrupt function execution.

galaxy/main/migrations/0108_sanitize_namespace_names.py
cutwater: +1

Issue #713, opened 2018-06-26
Annoying page reload when trying to remove users from namespace


newswangerd:
The x next to the owners name in the namespace detail view causes the page to reload when trying to remove the namespace's primary user.

Issue #703, opened 2018-06-22
Prevent non-admins from modifying Namespace name


chouseknecht:
A non-admin should not be allowed to change the *name* attribute on an existing Namespace.

- [x] In the UI, when editing an existing Namespace, disable the name field for non-admins
- [x] In the API, make sure the view does not allow a non-admin to update the name attribute for an existing Namespace object.

PR #715, opened 2018-06-26
Prevent users from changing namespace names

newswangerd:
Fixes: #703 
Fixes: #713


galaxyui/src/app/my-content/namespace-detail/namespace-detail.component.html
```diff
@@ -3,13 +3,24 @@
     <form [formGroup]="namespaceForm" novalidate>
         <div class="form-group">
             <label for="nameInput">Name <span class="required">*</span></label>
-            <input type="text"
+            <input *ngIf="!me.staff"
```
chouseknecht: I don't this is right. Think the formControl object should be disabled in the Javascript. See https://github.com/ansible/galaxy/blob/release/3.0/galaxyui/src/app/my-content/namespace-detail/namespace-detail.component.ts#L273

Issue #689, opened 2018-06-20
New user creation fails with a 500 Error


newswangerd:
I created a new user on github (https://github.com/newswangerd-test) and tried to sign up for an account on galaxy-qa, and I got this error. I think that the problem had to do with not setting up a name on my GitHub account, because adding my name in github seems to have solved the problem.

```
2018-06-20 16:13:28,468 ERROR exception: Internal Server Error: /accounts/github/login/callback/
Traceback (most recent call last):
  File "/var/lib/galaxy/venv/lib/python2.7/site-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/var/lib/galaxy/venv/lib/python2.7/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
    response = self._get_response(request)
  File "/var/lib/galaxy/venv/lib/python2.7/site-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/var/lib/galaxy/venv/lib/python2.7/site-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/var/lib/galaxy/venv/lib/python2.7/site-packages/allauth/socialaccount/providers/oauth2/views.py", line 73, in view
    return self.dispatch(request, *args, **kwargs)
  File "/var/lib/galaxy/venv/lib/python2.7/site-packages/allauth/socialaccount/providers/oauth2/views.py", line 143, in dispatch
    return complete_social_login(request, login)
  File "/var/lib/galaxy/venv/lib/python2.7/site-packages/allauth/socialaccount/helpers.py", line 153, in complete_social_login
    return _complete_social_login(request, sociallogin)
  File "/var/lib/galaxy/venv/lib/python2.7/site-packages/allauth/socialaccount/helpers.py", line 168, in _complete_social_login
    ret = _login_social_account(request, sociallogin)
  File "/var/lib/galaxy/venv/lib/python2.7/site-packages/allauth/socialaccount/helpers.py", line 55, in _login_social_account
    signal_kwargs={"sociallogin": sociallogin})
  File "/var/lib/galaxy/venv/lib/python2.7/site-packages/allauth/account/utils.py", line 165, in perform_login
    **signal_kwargs)
  File "/var/lib/galaxy/venv/lib/python2.7/site-packages/django/dispatch/dispatcher.py", line 193, in send
[...]

PR #716, opened 2018-06-26
Fix error during account creation if user has not given a name in github

newswangerd:
Fixes: #689


galaxy/main/signals/handlers.py
```diff
@@ -61,6 +61,9 @@ def user_logged_in_handler(request, user, **kwargs):
 
             login = social.extra_data.get('login')
             name = social.extra_data.get('name')
```
cutwater:
```
social.extra_data.get('name') or login
```

Issue #582, opened 2018-05-25
Check and improve UI performance


cutwater:
##### SUMMARY
According to users feedback our UI on Galaxy QA site is slow and browser seems to intensively load CPU.

We should profile UI performance and resource utilization and improve it if possible.

drybjed:
I'm seeing slowdown on Galaxy pages that display lots of imported repositories/content (currently ~134 things imported). This happens on Debian Jessie with Chromium 57, on Intel i7 CPU with 8 GB of RAM. Whenever I click anything on the page, the tab becomes unresponsive for about 30s-1m.

chouseknecht:
Had the same experience on My Content page when there's a large number of repositories. Reworked the Action menus to use components in PR #603 , which seemed to help. Will test more, after that code gets merged and deployed to QA.

chouseknecht:
Still work to do on My Content, regarding polling.

chouseknecht:
Changing this to enhancement, because I'm not sure we want to immediately back port it to v3.0

newswangerd:
There are noticeable slowdowns in UI performance on the My Content page for authors that have lots of imported repos. The issue seems to be caused by the import and delete buttons in the in the repository list

![screen shot 2018-07-06 at 10 32 46 am](https://user-images.githubusercontent.com/6063371/42384455-03552f6e-8108-11e8-9da9-70cb235ee7de.png)

The buttons are implemented as a separate component in angular and as such change detection has to be run on all of them whenever something happens on the page. Disabling change detection doubles the UI responsiveness, but something about rendering all the dropdown menus with the delete button is still causing significant lag whenever any UI actions are triggered. The UI here really needs to be rewritten to include pagination and searching so that we don't have to render so many UI elements.

chouseknecht:
Agree that pagination and searching is in order.

The buttons/menus are in components in order to make them dynamic.

PR #739, opened 2018-06-28
Implement lazy loading

newswangerd:
Configures Angular to load the search, my content and my imports pages in the background after the initial page load. This should decrease the time it takes for the app to load on the initial page request.

Part of a series of changes to address #582


chouseknecht:
This seems like a big change to the Javascript right before we go live. Would prefer to hold off for now.

galaxyui/src/app/my-content/my-content.routing.module.ts
```diff
@@ -36,7 +36,7 @@ const myContentRoutes: Routes = [
         canActivate: [AuthService]
```
chouseknecht: Seems like canActive should be a part of the app route module.

newswangerd: This isn't technically necessary because `canActivate` still protects the routes at the module level. I can add `canLoad` in `app-route.module` which will simply prevent the module from getting loaded at all if the user isn't logged in.

PR #758, opened 2018-06-29
Fix sanitized content names

cutwater:


brtknr:
What is the reason for this change? Is there an explanation anywhere? All our roles now suddenly have an underscore in them which is breaking the dependency list in requirements.yml! This is an unacceptable change.

Issue #776, opened 2018-07-01
Latest version not updated after Travis build notifies Galaxy


elliotweiser:
##### SUMMARY

There's a possible issue related to importing Ansible roles (and creating new versions/releases) through travis-ci.org using the [Notifications API](https://galaxy.ansible.com/api/v1/notifications)

##### STEPS TO REPRODUCE

I'm not entirely certain about how to reproduce this. I suppose you could try to create a new tag on an existing role and see if this imports correctly when notifying Galaxy from Travis CI. In any case, here's the relevant information in this case.

This is the [notification](https://galaxy.ansible.com/api/v1/notifications/198419/) received by Ansible Galaxy after [this Travis build](https://travis-ci.org/geerlingguy/ansible-role-homebrew/builds/398724591) passed. 

##### EXPECTED RESULTS

The notification should have triggered an import, which would have created the 2.4.3 release.

##### ACTUAL RESULTS

Nothing happened. 2.4.2 remains the latest release.

See versions in https://galaxy.ansible.com/geerlingguy/homebrew

chouseknecht:
Looks like it started the import, but there was an error. Here's a link to the import log:

https://galaxy.ansible.com/my-imports?namespace=geerlingguy&repository_name=homebrew&selected=257923&page_size=10

Not sure why it got a duplicate key error though. Investigating...

chouseknecht:
Somehow I think we're creating new content records, and the new content record of course has '-' converted to '_'.

The unique key on Content is: 'namespace', 'repository', 'name', 'content_type'

During import, in `galaxy/worker/importers/base`, the the get_or_create() is:

```
obj, is_created = models.Content.objects.get_or_create(
            namespace=ns,
            repository=repo,
            content_type=models.ContentType.get(self.data.content_type),
            original_name=original_name,
            defaults={
                'name': self.translate_content_name(name),
                'is_valid': False,
            }
```

I think our unique index is incorrect. It should be based on original_name, rather than name, and then the two would line up.

chouseknecht:
Fixed deployed to https://galaxy-qa.ansible.com - our public QA site. An import in QA should work now, and it should not change '-' to '_'.

chouseknecht:
The patch will be deployed to production tomorrow morning, btw.

lae:
@chouseknecht will this be a retroactive fix as well (i.e. will travis_lxc get renamed back to travis-lxc?)?

cutwater:
[...]

PR #784, opened 2018-07-02
Fix Content Creation Error

chouseknecht:
Proposed fix for #775 and #776

If merged, will need to be cherry-picked to 3.0.1

### DIFF
--- a/galaxy/worker/importers/base.py
+++ b/galaxy/worker/importers/base.py
@@ -69,10 +69,11 @@ class ContentImporter(object):
         try:
             # Check for an existing Content object matching name
             is_created = False
-            obj = models.Content.objects.get(namespace=ns,
-                                             repository=repo,
-                                             content_type=models.ContentType.get(self.data.content_type),
-                                             name=name)
+            obj = models.Content.objects.get(
+                namespace=ns,
+                repository=repo,
+                content_type=models.ContentType.get(self.data.content_type),
+                name=name)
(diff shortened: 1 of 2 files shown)


galaxy/worker/importers/base.py
```diff
@@ -65,17 +66,26 @@ def make_content(self):
...
-            defaults={
-                'name': self.translate_content_name(name),
-                'is_valid': False,
-            }
-        )
-
+        try:
+            # Check for an existing Content object matching name
+            is_created = False
+            obj = models.Content.objects.get(namespace=ns,
```
cutwater: Please try to fit your code to 80 characters line width.

galaxy/worker/importers/base.py
```diff
@@ -101,7 +111,7 @@ def _update_readme(self, content):
 
     def _log_create_content(self, content_id, is_created):
         action = 'Created new' if is_created else 'Found'
-        self.log.debug(
+        LOG.debug(
```
cutwater: Why this change?

chouseknecht: Bad. Will revert that. It's what I had to do in order to see the output. The self.log.debug() messages don't surface in the API response for some reason, and thus don't appear on the UI.

galaxy/worker/importers/base.py
chouseknecht: OK.

PR #833, opened 2018-07-05
Fix search links and icon sizing

### LINKED ISSUE #720: UI: Search page style fixes
From @trahman73 

Results Table
- Reduce size of icons text of Downloads, Watchers, etc. to match other details page.

### LINKED ISSUE #744: Open in new window does not work for next galaxy
##### STEPS TO REPRODUCE

Go to https://galaxy-qa.ansible.com/search?platforms=EL&order_by=-relevance&page_size=100

Right click on role - no option to open in new window.
CMD+click -> role opens in the same window.

##### EXPECTED RESULTS

it's possible to open role/user details and other stuff in new tab.

This is very important because galaxy is flooded with forks of the same role and you often want to compare the most popular, drop those which are just duplicates and then compare difference.

##### ACTUAL RESULTS

```

```

### LINKED ISSUE #812: Tiles are not clickable, but highlighted
##### SUMMARY

image

When I mousever `.list-pf-item`, it changes its background.

image

So I expect that Item is tile and interactive. But it's not interacive and click does nothing.
Instead, I should search for a clickable element.

After I mouseover `.description` the cursor changes to `pointing finger`, and I expect that I can click it. But again nothing happens.

In the end I found that only clickable things are build status and title

##### ACTUAL RESULTS

```

```

chouseknecht:
Fixes #744 - It's now clear where to click to get to content detail, links are links (as opposed to JS functions),  and right clicking shows 'Open Link in New Tab' and 'Open Link in New Window' options.

Fixes #720 - Icon size, count number size, and count label size are now consistent between search and content detail views.

Fixes #812 - No longer changes cursor to pointer when hovering over description. Remove background color change when hovering over a list item.

### DIFF
--- a/galaxyui/src/app/content-detail/repository/repository.component.html
+++ b/galaxyui/src/app/content-detail/repository/repository.component.html
@@ -18,34 +18,34 @@
     <div class="col-sm-8">
         <div class="repository-details">
             <div class="repo-count">
-                <i class="fa fa-download fa-2x"></i>
+                <i class="fa fa-download"></i>
                 <span class="count">{{ repositoryView.downloadCount }}</span>
-                <span class="count-label">Downloads</span>
+                <div class="count-label">Downloads</div>
             </div>
             <div class="repo-count">
-                <i class="fa fa-eye fa-2x"></i>
+                <i class="fa fa-eye"></i>
                 <span class="count">{{ repositoryView.watchersCount }}</span>
-                <span class="count-label">Watchers</span>
+                <div class="count-label">Watchers</div>
             </div>
             <div class="repo-count">
-                <i class="fa fa-star fa-2x"></i>
+                <i class="fa fa-star"></i>
                 <span class="count">{{ repositoryView.stargazersCount }}</span>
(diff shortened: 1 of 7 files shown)


galaxyui/src/app/content-detail/repository/repository.component.html
```diff
@@ -1,53 +1,65 @@
...
-                   [href]=""
-                   (click)="viewAuthor()"
-                   tooltip="View the Author's page">
-                   <img [src]="repositoryView.avatarUrl" > {{ repositoryView.displayName }}</a>
-            </div>
+
+    <div class="col-lg-2">
+        <div class="namespace-img">
+            <img class="img-wrapper" tooltip="View Author's Profile" [src]="namespace.avatar_url" (click)="viewAuthor()">
+            <a (click)="viewAuthor()"><span class="fa fa-user"></span> {{ namespace.name }}</a>
```
newswangerd: We need to fix these links as well, but I think that would be easier to do in a separate PR.

PR #845, opened 2018-07-07
Limit Repository char fields to 256 chars

### LINKED ISSUE #843: Error! Save repository failed: (Ensure this field has no more than 256 characters.)
##### SUMMARY
Trying to import a new repo and it is generating the following message:

```
Error! Save repository failed: (Ensure this field has no more than 256 characters.)
```

##### STEPS TO REPRODUCE
Import the repo from `https://github.com/privateip/ios`

##### EXPECTED RESULTS
The repo would be successfully imported

##### ACTUAL RESULTS
The web UI produced the error message and the repo was not imported

chouseknecht:
There was a special hook for admins that bypassed the security check and copying of GitHub data. Not sure why it existed. Maybe left over from the way imports originally worked. Removed, and then I was able to reproduce the error on #843

Now truncates GitHub char fields to 256 chars.

Fixes #843

### DIFF
--- a/galaxy/api/views/repository.py
+++ b/galaxy/api/views/repository.py
@@ -18,6 +18,8 @@
 import logging
 import re
 
+from six import string_types
+
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import Count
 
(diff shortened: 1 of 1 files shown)


galaxy/api/views/repository.py
```diff
@@ -115,14 +117,18 @@ def post(self, request, *args, **kwargs):
...
-                data[field] = repo[field]
+        repo = get_repo(provider_namespace, request.user, original_name)
+        if not repo:
+            raise PermissionDenied(
+                "User does not have access to {0}/{1} in "
+                "GitHub".format(provider_namespace.name, original_name))
+
+        for field in GITHUB_REPO_FIELDS:
+            if data.get(field) is not None:
+                if isinstance(data[field], string_types):
```
cutwater: Shouldn't it be `isinstance(repo[field], string_types)`?

chouseknecht: Yeah, that's right. It should be repo[field].

galaxy/api/views/repository.py
```diff
@@ -115,14 +117,18 @@ def post(self, request, *args, **kwargs):
...
-            for field in GITHUB_REPO_FIELDS:
-                data[field] = repo[field]
+        repo = get_repo(provider_namespace, request.user, original_name)
+        if not repo:
+            raise PermissionDenied(
+                "User does not have access to {0}/{1} in "
+                "GitHub".format(provider_namespace.name, original_name))
+
+        for field in GITHUB_REPO_FIELDS:
+            if data.get(field) is not None:
```
cutwater: Probably it should be `repo.get(field)`

Issue #860, opened 2018-07-11
JS lint "recommended" 


chouseknecht:
Add `"extends": "tslint:recommended",` to `tslint.json`, and fix the resulting JS errors. Most of the errors can be addressed by then running `ts lint galaxyui --fix` from within the `galaxyui` folder.

akaRem:
I've made some changes but there are some rules to discuss (probably).
Currently these rules are muted:
```
    "align": false,
    "array-type": false,
    "arrow-parens": false,
    "ban-types": false,
    "interface-name": false,
    "jsdoc-format": false,
    "max-classes-per-file": false,
    "no-consecutive-blank-lines": false,
    "object-literal-key-quotes": false,
    "object-literal-shorthand": false,
    "prefer-for-of": false,
    "trailing-comma": false,
```
You can read more about each of them on TSlint site [1]

I also didn't include rule for unused variables.

From my POV stricter linter is better in most of cases. But team may have another view on this.

The questions are:
1) Should I fix unused imports?
2) Which rules to enable?
3) There are some overrides like max-line-length, whitespace and others.. Should I drop them in favour of recommended or not?

If PR is Ok, then after it's merged, I will probably be able to add multiple PRs for different rules which will not conflict with each other and allow easier review and decision making process.

Full list of recommended rules is available in crsp repo [2]

[1] https://palantir.github.io/tslint/rules/
[2] https://github.com/palantir/tslint/blob/master/src/configs/recommended.ts

newswangerd:
@akaRem Nice work! I approved #866 and I think it's a pretty good starting point.

To answer your questions:
1. We definitely want to check for unused imports/variables
2. I would enable Align. The rest that you've disabled seem reasonable to me.
3. I would prefer to leave the current overrides as they are.

akaRem:
@newswangerd ok, I'll add these checks too.
I think it's better to do it with separate PR because that PR is already big enough.

I will probably propose other PRs with enabled checks plus fixes so team may look and decide which format looks more preferred.

akaRem:
Related PRs: https://github.com/ansible/galaxy/pull/866 https://github.com/ansible/galaxy/pull/890 https://github.com/ansible/galaxy/pull/891 https://github.com/ansible/galaxy/pull/892 https://github.com/ansible/galaxy/pull/893

newswangerd:
@akaRem #891 needs to be re-based before I can merge it. Also, see my comment on #893.

Thanks for your help!

[...]

PR #866, opened 2018-07-11
Fix/tslint rules

akaRem:
This PR adds "extends tslint:recommended" in tslint config for 
galaxyui according to https://github.com/ansible/galaxy/issues/860 and 
removes custom rules that duplicate provided config.

Recommended rules included in bundle that produce errors on 
`npm run lint` are mostly disabled, but a couple of fixes are already 
included.

This is initial PR with partially done work.
Work is done partially for easier review process and rebases when
required.

### DIFF
--- a/galaxyui/src/app/app-routing.module.ts
+++ b/galaxyui/src/app/app-routing.module.ts
@@ -1,13 +1,13 @@
 import { NgModule }              from '@angular/core';
-import { NotFoundComponent }     from './exception-pages/not-found/not-found.component';
+import { AuthService }           from './auth/auth.service';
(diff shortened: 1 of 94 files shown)


chouseknecht:
@akaRem 

Thank you for doing all this work. Really appreciate it!

@newswangerd, please review as soon as possible.

akaRem:
@chouseknecht @newswangerd Could you please take a look on https://github.com/ansible/galaxy/issues/860#issuecomment-404307958 ?
I can complete this work, but it's good to have consolidated view on which rules should be applied.

PR #880, opened 2018-07-13
Add regex to check if dates are valid before passing to moment.

newswangerd:


galaxyui/src/app/search/search.component.ts
```diff
@@ -483,10 +483,12 @@ export class SearchComponent implements OnInit, AfterViewInit {
     }
 
     private prepareContent(data: Content[], count: number) {
+        const datePattern = /^\d{4}.*$/;
         data.forEach(item => {
+            console.log(item.imported);
```
akaRem: Is it debug line that you forgot to remove?

PR #883, opened 2018-07-14
Add tests for CustomUser model

akaRem:
This patch introduces basic tests for CustomUser model and covers:
* model creation
* model validation
* some of model methods

### DIFF
--- a/galaxy/accounts/models.py
+++ b/galaxy/accounts/models.py
@@ -55,6 +55,7 @@ class CustomUser(auth_models.AbstractBaseUser,
                                       'invalid')
         ])
     full_name = models.CharField(_('full name'), max_length=254, blank=True)
+    # FIXME: This field looks unused
     short_name = models.CharField(_('short name'), max_length=30, blank=True)
     email = models.EmailField(_('email address'), max_length=254, unique=True)
     is_staff = models.BooleanField(
@@ -90,9 +91,11 @@ class CustomUser(auth_models.AbstractBaseUser,
     def __str__(self):
         return self.username
 
+    # FIXME: replace with django.core.urlresolvers.reverse(..)
     def get_absolute_url(self):
         return "/users/%s/" % urlquote(self.username)
 
+    # FIXME: This method looks unused
     def get_full_name(self):
         """
         Returns the first_name plus the last_name, with a space in between.
@@ -100,19 +103,23 @@ class CustomUser(auth_models.AbstractBaseUser,
         full_name = self.full_name
         return full_name.strip()
 
+    # FIXME: This method looks unused
     def get_short_name(self):
         """Returns the short name for the user."""
         return self.short_name.strip()
 
+    # FIXME: This method looks unused
     def get_num_roles(self):
         return self.roles.filter(active=True, is_valid=True).count()
 
+    # FIXME: This method looks unused
     def email_user(self, subject, message, from_email=None):
         """
         Sends an email to this User.
         """
         send_mail(subject, message, from_email, [self.email])
 
+    # FIXME: This method looks unused
     def hasattr(self, attr):
         return hasattr(self, attr)
 
@@ -130,6 +137,7 @@ class CustomUser(auth_models.AbstractBaseUser,
             'github_repo': g.repository.github_repo,
         } for g in self.starred.select_related('repository').all()]
 
+    # FIXME: This method looks unused
     def get_subscriber(self, github_user, github_repo):
         try:
             return self.subscriptions.get(
@@ -138,6 +146,7 @@ class CustomUser(auth_models.AbstractBaseUser,
         except exceptions.ObjectDoesNotExist:
             return None
 
+    # FIXME: This method looks unused
     def get_stargazer(self, github_user, github_repo):
         try:
             star = self.starred.get(
@@ -147,6 +156,7 @@ class CustomUser(auth_models.AbstractBaseUser,
         except exceptions.ObjectDoesNotExist:
             return None
 
+    # FIXME: This method looks unused
     def is_connected_to_github(self):
         connected = False
         for account in self.socialaccount_set.all():
--- a/galaxy/accounts/tests/__init__.py
+++ b/galaxy/accounts/tests/__init__.py
@@ -1,5 +1,4 @@
-# (c) 2012-2018, Ansible by Red Hat
-#
+# (c) 2012-2018, Ansible
 # This file is part of Ansible Galaxy
 #
 # Ansible Galaxy is free software: you can redistribute it and/or modify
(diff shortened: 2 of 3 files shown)


galaxy/accounts/models.py
```diff
@@ -68,7 +69,9 @@ class CustomUser(auth_models.AbstractBaseUser,
         db_index=True,
         help_text=_('Designates whether this user should be treated as '
                     'active. Unselect this instead of deleting accounts.'))
-    date_joined = models.DateTimeField(_('date joined'), default=timezone.now)
+    date_joined = models.DateTimeField(
+        _('date joined'),
+        default=lambda: timezone.now())
```
cutwater: Any reason for replacing function reference with lambda?

galaxy/accounts/tests/__init__.py
```diff
@@ -1,5 +1,3 @@
-# (c) 2012-2018, Ansible by Red Hat
```
cutwater: What is the reason for copyright removal?

galaxy/accounts/tests/test_custom_user_model.py
```diff
@@ -0,0 +1,374 @@
+# -*- coding: utf-8 -*-
```
cutwater: Missing copyright string as a part of Apache-2.0 license header.

galaxy/accounts/tests/test_custom_user_model.py
```diff
@@ -0,0 +1,374 @@
...
+from django.core.exceptions import ValidationError
+from django.contrib.auth.models import UserManager
+from django.db.utils import DataError, IntegrityError
+from django.utils import timezone
+from django.test import TestCase
+
+import mock
+import pytest
+
+from ..models import CustomUser
```
cutwater: Please prefer absolute module imports.

galaxy/accounts/models.py
```diff
@@ -90,29 +93,37 @@ class Meta:
...
+
+    # FIXME: This method looks unused
     def email_user(self, subject, message, from_email=None):
         """
         Sends an email to this User.
         """
-        send_mail(subject, message, from_email, [self.email])
+        self._send_email(subject, message, from_email, [self.email])
 
+    # FIXME: This method looks unused
```
cutwater: Looks like this method was left by predecessors. I thought I removed it.
Definitely not needed.

galaxy/accounts/models.py
```diff
@@ -90,29 +93,37 @@ class Meta:
...
     def get_short_name(self):
         """Returns the short name for the user."""
         return self.short_name.strip()
 
+    # FIXME: This method looks unused
     def get_num_roles(self):
         return self.roles.filter(active=True, is_valid=True).count()
 
+    _send_email = staticmethod(send_mail)  # for testing purposes
```
cutwater: Why not mocking `send_email` function or `email_user` method?

galaxy/accounts/tests/__init__.py
akaRem: In this case git decided that I moved file, but it's new actually.

galaxy/accounts/models.py
akaRem: **Yes.**
If you set `default=timezone.now` ref for `now` is stored in the model. Stubbing `timezone.now` will not work. And if I want to stub it, then it will be super tricky. 
On the other hand, `lambda: timezone.now()` encloses only `timezone`. In this case stubbing `timezone.now` will work perfectly.

galaxy/accounts/tests/test_custom_user_model.py
akaRem: Yes, and I have a question.
I'm not Red Hat employee. Should I put Red Hat copyright in headers — `(c) 2012-2018, Ansible by Red Hat`? Or may I just write `(c) 2012-2018, Ansible`?

akaRem: I will fix it, but I'm not sure what shall I write here.

galaxy/accounts/models.py
akaRem: Here we have the same problem as with `datetime.now`. `send_mail` ref is enclosed and you can not mock it. Well, technically I can do this, but you don't want to see this 😄
Ok, I will rework it in other way.

galaxy/accounts/models.py
akaRem: Could you please walk through all these labels and comment, what's really needed and what's not.

galaxy/accounts/tests/test_custom_user_model.py
cutwater: Leave it as is for now. We need to clarify our policies regarding copyrights.

galaxy/accounts/models.py
```diff
@@ -90,29 +93,37 @@ class Meta:
     def __str__(self):
         return self.username
 
+    # FIXME: replace with django.core.urlresolvers.reverse(..)
```
cutwater: Agree

galaxy/accounts/models.py
```diff
@@ -90,29 +93,37 @@ class Meta:
...
     def get_absolute_url(self):
         return "/users/%s/" % urlquote(self.username)
 
+    # FIXME: This method looks unused
```
cutwater: Agree. Method itself seems harmful.

galaxy/accounts/models.py
```diff
@@ -90,29 +93,37 @@ class Meta:
...
     def get_full_name(self):
         """
         Returns the first_name plus the last_name, with a space in between.
         """
         full_name = self.full_name
         return full_name.strip()
 
+    # FIXME: This method looks unused
```
cutwater: Agree.

galaxy/accounts/models.py
```diff
@@ -90,29 +93,37 @@ class Meta:
...
     def get_short_name(self):
         """Returns the short name for the user."""
         return self.short_name.strip()
 
+    # FIXME: This method looks unused
```
cutwater: Agree.

galaxy/accounts/models.py
```diff
@@ -130,6 +141,7 @@ def get_starred(self):
             'github_repo': g.repository.github_repo,
         } for g in self.starred.select_related('repository').all()]
 
+    # FIXME: This method looks unused
```
cutwater: Confirmed.

galaxy/accounts/models.py
```diff
@@ -138,6 +150,7 @@ def get_subscriber(self, github_user, github_repo):
         except exceptions.ObjectDoesNotExist:
             return None
 
+    # FIXME: This method looks unused
```
cutwater: Confirmed.

galaxy/accounts/models.py
```diff
@@ -147,6 +160,7 @@ def get_stargazer(self, github_user, github_repo):
...
```
cutwater: Confirmed.

galaxy/accounts/tests/test_custom_user_model.py
akaRem: @cutwater So for this and other PRs from my side I will add banners **without**
```
# (c) 2012-2018, <whatever>
``` 
line. Correct?

cutwater: Yes, we can fix it later.

galaxy/accounts/models.py
cutwater: It seems reasonable to me. Please make sure though that this change doesn't affect migrations, otherwise an appropriate migration should be created.

PR #884, opened 2018-07-14
Move common tests into common folder

akaRem:
Currently all tests in galaxy are located inside app folder. And only
common tests are exceptional.

This patch moves tests/unit files into common/tests to have consistent
test locations and easier searching.


akaRem:
@cutwater Django recommends to put tests alongside with app code (app here means main, common, api, ..). And it creates stubs for tests inside apps (see this codebase).
So I'd suggest to follow upstream recommendations.

cutwater:
@akaRem Acknowledged.

PR #888, opened 2018-07-14
Fix tripple quotes to match PEP

akaRem:
Tripple-doubleqoutes (`"""`) are used almost everewhere in this project.
And they are PEP compliant. However there are few places where
tripple-singlequotes (`'''`) are still used.

This patch replaces quotes with correct ones (`'''` -> `"""`)


cutwater:
@akaRem there is a type in your PR comment.

> This patch replaces quotes with correct ones (`"""` -> `'''`)

Probably should be: (`'''` -> `"""`)

PR #893, opened 2018-07-14
Enable array-type rule

akaRem:


akaRem:
@newswangerd It mostly about visual scanning. 

Sometimes it really helps.
``` TypeScript
Row<(string | number | null)[]>
```
Is it array of nulls or array of variants? Or array of rows? One symbol matters.
compare with
``` TypeScript
Row<Array<string | number | null>>
```

PS: This is just proposal. I can live without this rule. 🙂

PR #899, opened 2018-07-15
Add extra test files patterns

akaRem:
Currently pytest ignores some test files (like tests.py).
This patch add them to list of tests so they won't be ignore in
future.


akaRem:
Requires https://github.com/ansible/galaxy/pull/898 to be merged before review

PR #900, opened 2018-07-15
Fix license in galaxyui package.json

akaRem:
galaxyui/package.json has MIT license, but all banners in this repo
state that license is Apache-2.

This patch fixes license field to be Apache-2.

### DIFF
--- a/galaxyui/package.json
+++ b/galaxyui/package.json
@@ -1,7 +1,7 @@
 {
   "name": "galaxyui",
   "version": "0.0.0",
-  "license": "MIT",
+  "license": "Apache-2.0",
   "scripts": {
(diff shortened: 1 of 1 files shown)


cutwater:
I think we're using `Apache 2.0` license for all source code. In that case `Apache 2.0` headers should be added to all significant parts of UI code.

akaRem:
@cutwater do you want me to add banners to all galaxyui files?

newswangerd:
I'm okay with adding the license to the banner as long as we can find a way to do it automatically and as long as the comments get removed during the compilation process so that we aren't wasting bandwidth and load time by serving up tons of redundant comments.

PR #905, opened 2018-07-16
Fix flake8 docstrings errors

kbugaienko:
Improve PEP8 compliance by using `flake8-docstrings` plugin

### DIFF
--- a/galaxy/importer/loaders/role.py
+++ b/galaxy/importer/loaders/role.py
@@ -224,8 +224,9 @@ class RoleLoader(base.BaseLoader):
     linters = (linters.Flake8Linter, linters.YamlLinter)
 
     def __init__(self, content_type, path, root, metadata_path, logger=None):
-        """
-        :param str path: Path to role directory within repository
+        """Initializes a `RoleLoader` object.
+
+        :param str path: Path to role directory within repository.
         """
         super(RoleLoader, self).__init__(
             content_type, path, root, logger=logger)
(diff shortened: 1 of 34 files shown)


galaxy/main/signals/handlers.py
```diff
@@ -80,7 +80,8 @@ def user_logged_in_handler(request, user, **kwargs):
 @receiver(post_save, sender=models.ImportTask)
 def import_task_post_save(sender, **kwargs):
     """
-    When a role is imported enable the role in the user's repository cache
+    When a role is imported enable the role
```
akaRem: afaik, docstring for method should contain verb in 1st place.

galaxy/importer/loaders/role.py
```diff
@@ -224,9 +224,7 @@ class RoleLoader(base.BaseLoader):
     linters = (linters.Flake8Linter, linters.YamlLinter)
 
     def __init__(self, content_type, path, root, metadata_path, logger=None):
-        """
-        :param str path: Path to role directory within repository
-        """
+        """:param str path: Path to role directory within repository."""
```
akaRem: This looks really strange. Should we add some title here?

akaRem:
I can't find this plugin added/configured in PR. 
Is it one-time operation? Or is it supposed to be added as auto check?

akaRem:
@kbugaienko I think it's very important to add such config. Then CI will be the single source of truth for flake8 compliance.

akaRem:
image

this will affect business logic too

PR #928, opened 2018-07-19
Add pagination and searching repositories on My Content page

newswangerd:
#582

#935

Improves functionality and performance of page by rendering fewer components.

screen shot 2018-07-19 at 11 43 38 am
screen shot 2018-07-19 at 11 47 24 am

Edit: Filter now disappears when there are less than 5 repos. Pagination disappears when there are no pages. Alignment is tighter.

### DIFF
--- a/galaxyui/src/app/my-content/namespace-list/content/repositories-content/repositories-content.component.html
+++ b/galaxyui/src/app/my-content/namespace-list/content/repositories-content/repositories-content.component.html
@@ -1,3 +1,11 @@
+<div class="toolbar" *ngIf="maxItems > 5"
+    [ngClass]="{'toolbar-empty': filterConfig.appliedFilters.length === 0,
+                'toolbar-with-search': filterConfig.appliedFilters.length > 0}">
+    <pfng-toolbar [config]="toolbarConfig"
+                  (onFilterChange)="filterChanged($event)">
+    </pfng-toolbar>
+</div>
+
 <div class="repos-container">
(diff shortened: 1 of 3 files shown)


akaRem:
I think `My content` title should be kept to keep UI consistent in each tabs.

newswangerd:
@akaRem Oh, the My Content title is still there, the second screen shot is just scrolled down so you can see the pagination bar.

chouseknecht:
@newswangerd 

Tried this out locally. Couple issues:

1. When clicking the next page button, or changing the page size, the EmptyState flashes. Can we hide the EmptyState and just show a white background with the spinner until resources actually returns?
2. Both the pagination widget and filter widgets disappears after searching. Here's the reproducer:
    1. Pull up the Openstack namespace
    2. Search for `ara`.
    3. It gets a single match, and now both the page widget and filter widget are gone.

    Maybe in this case, the filter and pagination widgets should always be displayed. That might be easier, and more consistent.

chouseknecht:
@newswangerd 

Clicking the next/previous page, the screen jumps all over the place.

Haven't looked at the code, but the moment you click the button, the list of repos goes to 0, the screen flashes, a moment later  search results arrive, and the screen expands. The transition needs to be smooth. 

Part of the solution might be how the list gets updated in javascript. Another possibility might involve setting and removing a CSS class and using a CSS transition to make the list fade in and out. Might also need to set a min-height on the area that holds the list.

Issue #919, opened 2018-07-18
UI: issues with search page pagination


megamorf:
##### SUMMARY
Returning from a search result to the search page breaks the pagination

##### STEPS TO REPRODUCE
1. Open: https://galaxy.ansible.com/search?platforms=SLES&order_by=-relevance&page_size=10
2. Click ">" in the pagination pane at the bottom
3. Click on last search result on the page (the role 'pip' in my case)
4. Click your browser's back button to return to your search results page
5. The pagination site/entries counter will be messed up
6. Click ">" in the pagination pane at the bottom until everything breaks entirely

##### EXPECTED RESULTS
I expect the pagination to work.

##### ACTUAL RESULTS
![image](https://user-images.githubusercontent.com/1732128/42885351-0bcf17d4-8aa1-11e8-8fa7-f9e030f5a176.png)

chouseknecht:
This may already be fixed in the release/3.0.3 branch. Will test today.

chouseknecht:
Tested in [QA](https://galaxy-qa.ansible.com), and cannot reproduce. Believe it's addressed by #863

megamorf:
Can reproduce in in QA, please reopen.
![grafik](https://user-images.githubusercontent.com/1732128/42968592-09be808e-8ba4-11e8-9818-e19fa6111ef3.png)

newswangerd:
This was a sneaky issue. Basically when you load an existing search from a URL such as https://galaxy.ansible.com/search?vendor=false&order_by=-relevance&page_size=10&page=2, the page number gets loaded from the `page=XX` URL parameter. By default URL parameters are strings, so when you click on the button to go to the next page it just appends 1 to the string instead of increasing the page count by one.

PR #940, opened 2018-07-20
Convert page URL parameters to integers on search page.

newswangerd:
#919


akaRem:
If someone will make mistake during copy-paste or editing url, `pageSize` may become `0`, `NaN` or too big. Could you add crsp checks and fixes to such incorrect data?

newswangerd:
Unfortunately we can't really validate pageSize because we don't know what the size of the page is until after the query is executed. I added some basic type validation and an error message if the page size is out of bounds.

PR #950, opened 2018-07-23
Remove TS unused variables

akaRem:
Currently there are a lot of unused variables in project and they are
mostly represented as unused imports. There are also some unused
variables in scope of module and function definitions as well as some
unused private fields of TS classes.

This patch enables no-unused-variable rule in tslint config.
Due to the fact that it's not possible to check only unused imports,
other rule warnings must be fixed too.
So this patch includes removal of all unused variables found in
function bodies and variables declared but unused in scope of module.

Since constructor interface for some classes probably has some
limitations and it's not safe to remove such vars because there are no
tests yet, these cases are just wrapped with tslint ignore statements
to be reviewed in the future separately.


akaRem:
I'm not sure about such cases:
``` typescript
    constructor(
        private route: ActivatedRoute,
        // tslint:disable-next-line:no-unused-variable
        private router: Router,
        private importsService: ImportsService,
        private location: Location,
        private authService: AuthService
    ) {}
```
As I remember, constructor args are declared positionally, so It's not possible to drop `router` here.

In all other cases, where unused variable is last arg in constructor, I'll remove it.

newswangerd:
@akaRem one of the cool things about Angular is that the framework can detect what services and objects the constructor needs and provides all of them automatically when the object is created. When you create a component, you just tell Angular what you need in the constructor and it figures out how to provide it.

https://angular.io/guide/dependency-injection-in-action#injectable-and-nested-service-dependencies

akaRem:
@newswangerd thanks for the link.
Updated and rebased

PR #995, opened 2018-08-01
[WIP]- Custom event logging solution

newswangerd:
Adds the following:

### LogEventDirective
A logger that can be attached to `<a>` and `<button>` tags like so with the `appLogEvent` attribute:

```
<a [appLogEvent]="copy" [href]="" (click)="copyToClipboard()" class="btn btn-default" tooltip="Copy to clipboard"><i class="fa fa-copy"></a>

<a appLogEvent href="/docs/">Help</a>

<button appLogEvent type="button" class="btn btn-default" [ngClass]="{'btn-primary': showingView === ViewTypes.detail}" tooltip="Return to main content view" (click)="toggleView(ViewTypes.detail)">Details</button>
```

`appLogEvent` can accept an option name attribute. If this is omitted, the value inside the tag is used instead.

Logging an `<a>` tag captures the elements `href=""` attribute. Logging a `<button>` tag just logs the button's name.

### EventLoggerService
Gets used by `LogEventDirective` to send data back to the back end. Can be used within component event handlers as well in situations where the `appLogEvent` attribute can't be added to the HTML template (such as is the case with patternfly components)

Link data gets sent to the API as:
```
{
  type: 'link',
  payload: {
    href: "https://www.galaxy.com/davidwittman/redis",
    name: "redis",
    current_page: "/search"
  }
}
``` 

Button data:
```
{
  type: 'button',
  payload: {
    name: "Read Me",
    current_page: "/davidwittman/redis"
  }
}
```

### Todo
- ~Add the following to SharedModule:~
   - ~clipboard component~
   - ~header component~
   - ~page loading component~

- Extend EventLoggerService to be able to log page load events and error events.

- ~Apply the `EventLoggerService` and `appLogEvent` to all links and buttons.~

- Aggregate events into a queu and send them all at once instead of calling the api each time someone clicks a link or a button.


newswangerd:
This is impossible to rebase.

PR #1000, opened 2018-08-01
Fix broken error handling

### LINKED ISSUE #981: UI error messages are broken when the backend returns a 500
The UI should return a coherent error message, but instead we get this garbage when a 500 happens on the server:

image

chouseknecht:
Fixes #981 

In the error object returned by the API, we assumed the `error` key pointed to an object, when sometimes it might actually point to a string.


cutwater:
@chouseknecht Shouldn't it be also fixed on server side so that backend returns consistent error messages?

chouseknecht:
@cutwater Possibly. This fixes the immediate problem though. If there's a more wholistic approach you think we should take to handling errors, lets discuss, and plan how to do it.

Issue #861, opened 2018-07-11
JS code formatting


chouseknecht:
VSCode has an extension called Beautify. It supports putting project code format settings in a `.jsbeautifyrc` file. The extension, if enabled, will then apply the settings either while a file is being edited, or on save.

- Does this extension also work in other popular IDEs?
- How do we enforce a JS formatting standard? There should be some automated check that runs against new PRs. There should be a way for a developer to test locally prior to PR submission.
- If the first 2 points can be solved, then develop a standard set of rules for the project.

akaRem:
There is also Prettier [1] extension for VSCode [2]
It's also very popular and it's shipped for many IDEs and editors (see editor support section on main site)

On the other hand, beautify plugin for VSCode [3] does not state that it supports TypeScript.

[1] https://prettier.io
[2] https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode 
[3] https://marketplace.visualstudio.com/items?itemName=HookyQR.beautify

newswangerd:
I've spent some time experimenting with prettier and jsbeautify and I've decided that using a code formatter doesn't really solve any problems for us. Our code is already pretty well formatted, and TSLint does a good job of enforcing the standards that we want to enforce. In some cases, Prettier makes the code harder to read and overall I think it will make accepting community contributions more difficult because it's an extra set of steps for new authors to follow.

PR #1004, opened 2018-08-02
Enforce TypeScript and Less code formatting with prettier.

newswangerd:
- Added `galaxyui/.prettierrc.yaml` to configure code formatting options
- Added `make dev/prettier` to format `.ts` and `.less` files in `galaxyui/src`
- Added prettier to node dependencies
- Added `make test/prettier` to check for unformatted code
- Added check during dev container start to verify that no unformatted code is added
- Updated contributing guide

#861

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -432,6 +432,26 @@ To run lint checks against JavaScript/TypeScript sources, execute:
 
     $ make dev/jslint
 
+Formatting your code
+^^^^^^^^^^^^^^^^^^^^
+
+We use prettier to enforce code formatting for all of our TypeScript and less files.
+To automatically format your Angular code run:
+
+.. code-block:: console
+
+    $ make dev/prettier
+
+Unformatted code will cause the travis build to fail when you push your changes to
+GitHub.
+
+It's recommended that you set up prettier on your editor if you're making lots of
+changes to anything in ``galaxyui/``. Prettier is supported by most major editors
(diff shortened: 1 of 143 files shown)


newswangerd:
@cutwater I added it to `scripts/docker/dev/start-develop.sh` because that's where `make test/jslint` is. Should the prettier and JSLint both be added to the travis file?

newswangerd:
We probably should move JS Lint out of the `start-develop.sh` because it makes it prevents the container from starting when there are linting errors, which is really annoying

chouseknecht:
@newswangerd if you take jslint out of the startup script, than you have to add it somewhere in the Travis script.

newswangerd:
@chouseknecht that's what I had in mind. I'm testing that now

CONTRIBUTING.rst
```diff
@@ -432,6 +432,19 @@ To run lint checks against JavaScript/TypeScript sources, execute:
...
+
+We use prettier to enforce code formatting for all of our TypeScript and less files.
+To automatically format your Angular code run:
+
+.. code-block:: console
+
+    $ make dev/prettier
+
+Unformatted code will cause the travis build to fail when you push your changes to
+GitHub.
```
chouseknecht: We need to mention that there is a `.prettierrc.yaml`, where settings live.

chouseknecht: Also, might be a good idea to add a link to https://prettier.io/docs/en/editors.html, so people have some idea of how to configure their editor.

PR #1022, opened 2018-08-03
Reformat some APB paramater metadata on save

djzager:
This makes it possible for the [Ansible Galaxy Adapter](https://github.com/openshift/ansible-service-broker/blob/master/docs/config.md#ansible-galaxy-registry) in the [broker](https://github.com/openshift/ansible-service-broker) can properly handle parameter fields when rendering an APB in the service-catalog UI (for OpenShift).

### DIFF
--- a/galaxy/importer/loaders/apb.py
+++ b/galaxy/importer/loaders/apb.py
@@ -33,6 +33,18 @@ class APBMetaParser(object):
     # Tags should contain lowercase letters and digits only
     TAG_REGEXP = re.compile('^[a-z0-9]+$')
 
+    # APB parameters should be in json-schema form
+    PARAM_KEY_MAP = {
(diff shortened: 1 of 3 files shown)


galaxy/importer/loaders/apb.py
```diff
@@ -137,6 +149,13 @@ def _check_plans(self):
                         expected_parameter_keys,
                         'plans[{0}].parameters[{1}]'.format(idx, pidx),
                         params)
+                    for param_key in params.keys():
```
djzager: I really had mixed feelings about adding this logic here as it's really doing validation of plans. My reasoning for including it here is two-fold: 1) easiest 2) most of the work to get down to the plans + parameters has already been done.

Issue #1008, opened 2018-08-03
A "Deprecated" flag and other visual cues


tima:
### Use Case

The ability to label a piece of contributed content that it should *not* be used going forward for any number of reasons. Rather than remove it from galaxy the contributor (or perhaps Galaxy admin) can flag it as deprecated so those who really want it can still get it and new consumers can be advised to look elsewhere.

### Proposed Solution

A simple clear marker on listings along with a search filter (deprecated entries should not be displayed by default) and perhaps "greying out" those search results if they are included as an additional visual cue. 

https://forge.puppet.com/rmueller/cron 
https://forge.puppet.com/rmueller (note cron listing)

### Alternatives

N/A

### Implementation

Covered in the explanation above.

chouseknecht:
Does the idea of *deprecated* apply at a Collection (i.e. Repository) level, or down at the content item level?

Can a user only deprecate something through the UI, or are there corresponding *deprecated* attributes that should be added to metadata?

This has implications for the Galaxy CLI, Mazer, as well. If a user tries to install a *deprecated* content item or Collection, then Mazer should throw a deprecation warning.

tima:
> Does the idea of deprecated apply at a Collection (i.e. Repository) level, or down at the content item level?

Start with the simplest thing and see what happens -- that seems like the whole collection/repo.

> Can a user only deprecate something through the UI, or are there corresponding deprecated attributes that should be added to metadata?

I think no. Too much replication. I think it wise to give the galaxy admin the ability to deprecate something like a role has a zero quality score and has been update in 3 years and we want to start cleaning up we can mark it deprecated.

> This has implications for the Galaxy CLI, Mazer, as well. If a user tries to install a deprecated content item or Collection, then Mazer should throw a deprecation warning.

Agreed. How that is done seems like an implementation detail.

chouseknecht:
@newswangerd 

Summarizing the above comments to a list of tasks...

- [ ] On My Content, in the kabob menu for a repository, add a 'Deprecate' option, and an 'Undeprecate' option. If a repository is not deprecated, the Deprecate option should be visible. If a repository is deprecated, the Undeprecate option should be visible.
- [ ] Both admins and namespace owners (contributors) will have access to the Deprecate and Undeprecate options.
- [ ] On the Repository model add a 'deprecated' attribute
- [ ] Add the 'deprecated' attribute to the Repository serializer
[...]

PR #1071, opened 2018-08-14
Added basic deprecation system.

newswangerd:
#1008

Adds an option to deprecate content from the My Content page:
deprecate-01

Defaults searches to only show non deprecated content:
deprecate-02

Adds tags and other visual cues to deprecated content on My Content, Search, Author and Content Detail pages.

Search:
screen shot 2018-08-14 at 11 24 49 am

Author:
screen shot 2018-08-14 at 11 25 30 am

Content Detail:
screen shot 2018-08-14 at 11 26 10 am


galaxy/api/views/repository.py
```diff
@@ -220,12 +231,16 @@ def update(self, request, *args, **kwargs):
...
-        import_task = tasks.create_import_task(repository, request.user)
-
         serializer = self.get_serializer(instance=instance)
         data = serializer.data
-        data['summary_fields']['latest_import'] = \
-            serializers.ImportTaskSerializer(import_task).data
+
+        # Don't create an import task if we're just updating the deprication
+        # status
+        if updated != ['deprecated']:
```
cutwater: Do we really need to create import task if other fields (e.g. `original_name`, `description`, etc.) got updated?

newswangerd: Which fields should kick off an import if they are update?

chouseknecht: Technically we don't need to run an import when a repo get's updated. However, it doesn't hurt anything, if it does run. Currently in the UI, it's not possible for a users to update repository attributes, so I think we can leave what David has for now and revisit later, if needed.

PR #1075, opened 2018-08-17
Email management API endpoints

chouseknecht:
The endpoints:
- View all emails, and create new email by posting to `/api/v1/emails`
- Update and Delete an emails at `/api/v1/emails/#/`
- View all emails for a user at `/api/v1/users/#/emails/`
- View user's primary email at `/api/v1/me/`

Security:
- An authenticated user can only update his/her own email addresses
- Non-admin can only view his/her own email addresses
- Anonymous users cannot view email addresses

Caveats:

There is no way to check that at least one email address is marked as primary, as I'm not sure where we would put that check. Wondering if we can check for a primary_email at `/api/v1/me/` when the user logs in, throw a warning with link/directions on how to fix.

### DIFF
--- a/galaxy/api/views/views.py
+++ b/galaxy/api/views/views.py
@@ -159,6 +159,7 @@ class ApiV1RootView(base_views.APIView):
         data['tags'] = reverse('api:tag_list')
         data['users'] = reverse('api:user_list')
         data['surveys'] = reverse('api:community_survey_list')
+        data['emails'] = reverse('api:email_list')
         return Response(data)
 
 
--- a/galaxy/api/access.py
+++ b/galaxy/api/access.py
@@ -19,6 +19,7 @@
 
 import logging
 
+from allauth.account.models import EmailAddress
 from django.contrib.auth import get_user_model
 from rest_framework.authtoken.models import Token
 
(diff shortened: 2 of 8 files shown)


chouseknecht:
@newswangerd 

How about, if you edit it, it becomes 'unverified'. That seems easier to implement.

chouseknecht:
@newswangerd 

Added a custom validator that prevents modifying a verified email address.

newswangerd:
Emails should definitely be verified. 

My proposal would be as follows (based on the UI in #1072):
1. User hits the verify button and receives a notification that an email has been sent.
screen shot 2018-08-21 at 8 50 13 am

2. The preferences page begins polling the API for email updates

3. The "Verify Email" button changes to "Resend Email" until the verification confirmation is received. 

4. Once the confirmation is received, stop polling the API, mark the email as verified, and disable the "Verify Email" button.

chouseknecht:
Removed WIP. We agreed to add email verification workflow as a separate PR.

galaxy/api/views/views.py
```diff
@@ -168,7 +168,11 @@ def get(self, request, format=None):
         data['search'] = reverse('api:search_view')
         data['tags'] = reverse('api:tag_list')
         data['users'] = reverse('api:user_list')
+<<<<<<< HEAD
```
cutwater: Merge conflict unresolved.

chouseknecht: Fixed.

Issue #1061, opened 2018-08-13
Add Content Rating and Feedback System.


newswangerd:
We want to add a short survey to content items in Galaxy that allow for users to rate content that they find on the site. The goal here is to make it easier for end users to determine which content bundles will work the best for them.

- [x] Design UI for community surveys.
- [x] Set up database models
- [x] Set up API
- [x] Add the survey to the content detail page.

PR #1082, opened 2018-08-20
Adds "community survey" to repos so that users can submit ratings.

newswangerd:
#1061

### Back End

- Adds `community_survey` to the database model
- Adds new API at `/api/v1/community_surveys/` for viewing/adding/updating surveys
- Adds `community_survey` field to the `Repository` model

### Front End
- Adds form on content detail page for submitting surveys
screen shot 2018-08-20 at 4 39 39 pm

screen shot 2018-08-22 at 11 21 12 am

### To Do
- ~Prevent users from rating their own content~ <- We're allowing this for now
- Add community survey score to search results/author page
- ~Make the form more responsive~


galaxy/api/access.py
```diff
@@ -330,6 +330,20 @@ def can_read(self, obj):
...
 
+class CommunitySurveyAccess(BaseAccess):
+    def can_read(self, obj):
+        return True
+
+    def can_add(self, data):
+        return self.user.is_authenticated()
+
+    def can_change(self, obj, data):
+        return self.user.is_authenticated()
```
chouseknecht: Should I only be allowed to modify a survey that I created? If so, think we can add that check here.

galaxy/api/views/survey.py
```diff
@@ -0,0 +1,126 @@
...
+        except ObjectDoesNotExist:
+            raise APIException(
+                'Repository does not exist.'
+            )
+
+        data['repository'] = repo
+        data['user'] = request.user
+
+        try:
+            survey = models.CommunitySurvey.objects.create(**data)
```
chouseknecht: `data` needs to be properly validated by the serializer. See http://www.django-rest-framework.org/api-guide/validators/. Guessing that `repo` can be handled in the serializer as well. May require a custom class validator.

galaxy/api/views/survey.py
```diff
@@ -0,0 +1,126 @@
...
+    def update(self, request, *args, **kwargs):
+        data = request.data
+        instance = self.get_object()
+
+        if (request.user.id != instance.user.id):
+            return Response(status=status.HTTP_403_FORBIDDEN)
+
+        for key in data:
+            if key in SURVEY_FIElDS:
+                setattr(instance, key, data[key])
```
chouseknecht: Same as above. We need to start following DRF, and use proper validation techniques, as much as possible.

newswangerd:
In order to make the form more responsive, the plan is to do some refactoring to the Patternfly menu so that it takes up less space, which will be handled in a different PR

galaxy/api/serializers/survey.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+
+    def get_url(self, obj):
+        if obj is None:
+            return ''
+        return reverse('api:community_survey_detail', args=(obj.pk,))
+
+    def get_summary_fields(self, instance):
+        return {
+            'community_score': instance.repository.community_score
```
chouseknecht: Why did we put *community_score* in *summary_fields*? Since it's a direct attribute of the Repository object, it should just be returned as a field. *Summary_fields* is intended for returning attributes from relationships or related objects.

chouseknecht:
One minor comment regarding the *survey* serializer. Other than that, LGTM.

galaxy/api/serializers/survey.py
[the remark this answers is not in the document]
newswangerd: Thanks for the feedback. I'll update this and merge.

Issue #1048, opened 2018-08-09
Content scoring - First iteration


chouseknecht:
Based on https://docs.google.com/document/d/1pOS_l0B_xz5MqLB3y6MxrnRbg4buWCpf7NdsKF-KTOE/edit?usp=sharing, we want to begin using [Ansible Lint]() to generate a content score. 

To achieve this we'll break the work into two or more iterations. The first iteration will encompass the following:

- Develop an understanding of the following: how to use Ansible Lint, how Ansible Lint analyzes the YAML files found in a role directory structure, how to write custom rules
- Copy rules from Ansible Lint. Any rules from [Ansible Lint](https://github.com/willthames/ansible-lint/tree/master/lib/ansiblelint/rules) that we want to use for analyzing and scoring Galaxy content, copy to [galaxy-lint-rules](https://github.com/ansible/galaxy-lint-rules), and assign a 'GALAXY*' id value. All rules that will be applied by Galaxy must live in [galaxy-lint-rules](https://github.com/ansible/galaxy-lint-rules), so that we can in the future associate them to our scoring algorithm and thoroughly document.
- Develop one or two custom rules in [galaxy-lint-rules](https://github.com/ansible/galaxy-lint-rules).
- Integrate Ansible Lint with the [Galaxy import process](https://github.com/ansible/galaxy/tree/devel/galaxy/importer), such that:
    - Ansible Lint executes against a traditional role, or each individual role in a multi-role repository
    - STDOUT/STDERR from Ansible Lint is returned by the API, and visible in the UI
    - Ansible Lint executes only the rules found in [galaxy-lint-rules](https://github.com/ansible/galaxy-lint-rules)

PR #1086, opened 2018-08-21
Added ansible-lint to importer, and scripts pull custom lint rules

awcrosby:
Adds ansible-lint output to importer using custom rules pulled from ansible/galaxy-lint-rules repo. Supports #1048

### DIFF
--- a/galaxy/importer/linters/__init__.py
+++ b/galaxy/importer/linters/__init__.py
@@ -72,3 +72,33 @@ class YamlLinter(BaseLinter):
         for line in proc.stdout:
             yield line.strip()
         proc.wait()
+
+
+class AnsibleLinter(BaseLinter):
+
+    cmd = 'ansible-lint'
+
+    def _check_files(self, paths):
+        rules_path = '/galaxy-lint-rules/rules'
+        cmd = [self.cmd, '-p', '-r', rules_path, '.']
+        logger.debug('CMD: ' + ' '.join(cmd))
+
+        # different logic needed for multi role repos since
+        # ansible-lint issue role path cannot contain '/'
+        cwd = (
+            self.root
+            if paths == ['.'] else
+            '/'.join((self.root, paths[0]))
+        )
+        proc = subprocess.Popen(cmd, cwd=cwd, stdout=subprocess.PIPE)
+
+        for line in proc.stdout:
+            line_list = line.split(' ')
+            rel_path = ['.'] + line_list[0].split('/')[3:]
+            line_list[0] = '/'.join(rel_path)
+            line = ' '.join(line_list)
+            yield line.strip()
+
+        # returncode 1 is app exception, 0 is no linter err, 2 is linter err
+        if proc.wait() not in (0, 2):
+            yield 'Exception running ansible-lint, could not complete linting'
--- a/galaxy/importer/tests/test_linters.py
+++ b/galaxy/importer/tests/test_linters.py
@@ -16,6 +16,9 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 import tempfile
+import os
+import shutil
+import pytest
 
 from galaxy.importer import linters
 
@@ -35,7 +38,7 @@ def test_flake8_ok():
 
         linter = linters.Flake8Linter()
         result = list(linter.check_files(fp.name))
-        assert result == []
+    assert result == []
 
 
 FLAKE8_TEST_FILE_FAIL = """
@@ -63,7 +66,7 @@ def test_flake8_fail():
             "{0}:7:1: W191 indentation contains tabs",
         ]
         expected = [s.format(fp.name) for s in expected]
-        assert sorted(result) == sorted(expected)
+    assert sorted(result) == sorted(expected)
 
 
 YAMLLINT_TEST_FILE_OK = """---
@@ -80,7 +83,7 @@ def test_yamllint_ok():
 
         linter = linters.YamlLinter()
         result = list(linter.check_files(fp.name))
-        assert result == []
+    assert result == []
 
 
 YAMLLINT_TEST_FILE_FAIL = """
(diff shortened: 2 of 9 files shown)


chouseknecht:
@awcrosby 

Tested this out. Here's output from the import of my role:

linting-outpout

Very cool! Nice work connecting Ansible Lint to the import process.

Couple of thoughts... 

1. Where we say 'Info [role] linting...'. Can we change 'linting...' to 'Running Ansible Lint', so that it's very clear what process is about to rung.
2. On the Error lines. Can we simplify the output a bit. Skinny it down to something more like:
    ```
    ERROR [role] meta/main.yml:19: [EGALAXYTEST201] Trailing whitespace
    ```

chouseknecht:
@awcrosby 

Will review this more before tomorrow. Haven't looked at the code. Just downloaded and ran to see the functionality.

galaxy/importer/linters/__init__.py
```diff
@@ -72,3 +72,29 @@ def _check_files(self, paths):
...
             yield line.strip()
         proc.wait()
+
+
+class AnsibleLinter(BaseLinter):
+
+    cmd = 'ansible-lint'
+
+    def _check_files(self, paths):
+        rules_folder = '/galaxy/galaxy/importer/galaxy-lint-rules/rules'
```
cutwater: 1. Hardcoded docker specific path.
2. Directory term is preferred to use.

galaxy/importer/linters/__init__.py
```diff
@@ -72,3 +72,29 @@ def _check_files(self, paths):
...
+        # ansible-lint issue role path cannot contain '/'
+        if paths == ['.']:  # single role repo
+            cmd = [self.cmd, '-p', '-r', rules_folder] + paths
+            logger.debug('CMD: ' + ' '.join(cmd))
+            proc = subprocess.Popen(cmd, cwd=self.root, stdout=subprocess.PIPE)
+        else:  # multi role repo
+            cmd = [self.cmd, '-p', '-r', rules_folder] + ['.']
+            logger.debug('CMD: ' + ' '.join(cmd))
+            cwd = '%s/%s' % (self.root, paths[0])
+            proc = subprocess.Popen(cmd, cwd=cwd, stdout=subprocess.PIPE)
```
cutwater: You can move `subprocess.Popen` and `logger.debug` outside if to reduce code duplication.

galaxy/importer/linters/__init__.py
```diff
@@ -72,3 +72,29 @@ def _check_files(self, paths):
...
+
+        for line in proc.stdout:
+            yield line.strip()
+        proc.wait()
+        if proc.returncode == 1:
```
cutwater: What if returncode is not `0` or `1`?

galaxy/importer/linters/__init__.py
```diff
@@ -72,3 +72,29 @@ def _check_files(self, paths):
...
+            yield 'Error running ansible-lint, could not complete linting'
```
cutwater: Exception?

galaxy/importer/linters/__init__.py
```diff
@@ -72,3 +72,29 @@ def _check_files(self, paths):
...
+            proc = subprocess.Popen(cmd, cwd=self.root, stdout=subprocess.PIPE)
+        else:  # multi role repo
+            cmd = [self.cmd, '-p', '-r', rules_folder] + ['.']
+            logger.debug('CMD: ' + ' '.join(cmd))
+            cwd = '%s/%s' % (self.root, paths[0])
+            proc = subprocess.Popen(cmd, cwd=cwd, stdout=subprocess.PIPE)
+
+        for line in proc.stdout:
+            yield line.strip()
+        proc.wait()
```
cutwater: 1. `.wait()` method returns a returncode already.

galaxy/importer/linters/__init__.py
webknjaz: Even more: only `cwd=` differs in these blocks.
(`if path == ['.']`, use `['.']`, else use same `['.']`, just not in var, for the last path part)

galaxy/importer/linters/__init__.py
```diff
@@ -72,3 +72,29 @@ def _check_files(self, paths):
...
+class AnsibleLinter(BaseLinter):
+
+    cmd = 'ansible-lint'
+
+    def _check_files(self, paths):
+        rules_folder = '/galaxy/galaxy/importer/galaxy-lint-rules/rules'
+
+        # different logic needed for multi role repos since
+        # ansible-lint issue role path cannot contain '/'
+        if paths == ['.']:  # single role repo
```
webknjaz: Refactor:
```python
cmd = [self.cmd, '-p', '-r', rules_folder, '.']
logger.debug('CMD: ' + ' '.join(cmd))
cwd = (
    self.root
    if paths == ['.'] else
    '/'.join((self.root, paths[0]))
)
proc = subprocess.Popen(cmd, cwd=cwd, stdout=subprocess.PIPE)
```

galaxy/importer/linters/__init__.py
webknjaz: 3. In general, I'd advise using pathlib as an underlying implementation of working with paths. (See pathlib2 for a good and improved backport from stdlib)

galaxy/importer/tests/test_linters.py
```diff
@@ -104,3 +106,87 @@ def test_yamllint_fail():
...
+ANSIBLELINT_TEST_FILE_OK = """---
+- name: Add mongodb repo apt_key
+  become: true
+  apt_key: keyserver=hkp
+"""
+
+
+def test_ansiblelint_ok():
+    try:
+        temp_root = tempfile.mkdtemp()
```
webknjaz: Use pytest fixtures for this, it doesn't belong to the test:
```python
import pytest
...
@pytest.fixture
def temp_root():
    try:
        tmp = tempfile.mkdtemp()
        yield tmp
    finally:
        shutil.rmtree(tmp)
...
def test_ansiblelint_ok(temp_root):
    ...
    assert stuff without any try/except/finally here
    ...
```

galaxy/importer/tests/test_linters.py
```diff
@@ -104,3 +106,87 @@ def test_yamllint_fail():
...
+        os.makedirs(full_path)
+        file_name = os.path.join(full_path, 'main.yml')
+        with open(file_name, 'w') as fp:
+            fp.write(ANSIBLELINT_TEST_FILE_OK)
+            fp.flush()
+
+            linter = linters.AnsibleLinter()
+            linter.root = '%s/%s' % (temp_root, role_path.split('/')[0])
+            result = list(linter.check_files(['.']))
+            assert result == []
```
webknjaz: better do assertions outside of with-block to avoid interfering with context managers

galaxy/importer/tests/test_linters.py
```diff
@@ -104,3 +106,87 @@ def test_yamllint_fail():
...
+
+
+def test_ansiblelint_ok():
+    try:
+        temp_root = tempfile.mkdtemp()
+        role_path = 'role/tasks'
+        full_path = os.path.join(temp_root, role_path)
+        os.makedirs(full_path)
+        file_name = os.path.join(full_path, 'main.yml')
+        with open(file_name, 'w') as fp:
```
webknjaz: This could be moved to fixture as well.

scripts/docker/dev/start-develop.sh
```diff
@@ -15,6 +15,17 @@ if [ "${TEST}" != "1" ]; then
     make migrate
 fi
 
+# clone or pull the repo: galaxy-lint-rules
```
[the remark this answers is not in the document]
awcrosby: 1. Moved clone to dev/Dockerfile and release/Dockerfile.build
2. Cloned to root rather than project source tree. Moved pull to command: make dev/lint-rule-update

galaxy/importer/linters/__init__.py
awcrosby: I started to use os.path to build a relative path, but then re-added hardcoded path since I moved repo location to root.

galaxy/importer/linters/__init__.py
awcrosby: Updated

galaxy/importer/linters/__init__.py
awcrosby: returncode = 1 is ansible-lint app exception, 0 is no linter error, 2 is linter error; clarified with a code comment.

galaxy/importer/linters/__init__.py
awcrosby: Renamed to exception. This summary message is displayed to the user instead of outputting stderr, because stderr might be difficult to read in our import process.

galaxy/importer/linters/__init__.py
awcrosby: Updated

galaxy/importer/linters/__init__.py
awcrosby: Updated, thanks!

galaxy/importer/tests/test_linters.py
awcrosby: Updated

galaxy/importer/tests/test_linters.py
awcrosby: Updated

galaxy/importer/tests/test_linters.py
awcrosby: Added duplication to function

Issue #1111, opened 2018-08-28
Search tracking


chouseknecht:
### Problem

We want to answer the following:

- What are users searching for?
- How many results are being returned for a given search?
- How many results are being returned on average?
- Are users clicking on a result found on the first page?
- Are users paging through results?
- How many searches result in no clicks?
- How many searches result in a click?
- Which content items get the most clicks?
- On average, how many searches does a user perform before clicking a result?
- On average, how many searches does a user perform during a visit to the site?

#1013 proposed a generalized API endpoint that can accept an event with a payload, which will work well for isolated events, but not necessarily for connecting events together.

To answer the above questions, events need to be tracked by user, date, and search. For example, the user may submit a search, click the next-page button multiple times, and then click a result. They may then click the browser back button to return to the search, and a click on a second result. Ideally, we want to capture the data in such a way that it’s easy to associate all of the clicks to a single search performed by a single user. And for the user and search, the next-page button was clicked X times, and 2 results were clicked.

The resulting data structure from all of the events captured in the above example should look something like the following:

```
{
    ‘session_key’: ‘xxxxxx’,
    ‘date’: 28-Aug-2018,
    ‘search_id’: ‘xxxx-xxxx-xxxx-xxxx’,
    ‘search_params’: {
        ‘tags’: [‘web’, ‘compose’],
        ‘platforms’: [‘ubuntu’],
        ‘keywords’: [‘php’]
        ‘contributor’: ‘’,
        ‘community’: true
        ‘deprecated’: false
    },
    'search_results': 111   # number of content items returned by the API
    ‘next_page_clicks’: 3,
    ‘prev_page_clicks’: 1,
    ‘results_clicked’: [24, 26],
    ‘repositories_viewed’: [3014, 20615],    # Repository IDs
    ‘Content_items_viewed’: [12123, 23490]    # Content IDs
}
```

### Proposed Solution

Represent the above data structure in PostgreSQL using Django ArrayField and JSONField types as needed. Store the data by unique *session_key* (i.e., user session cookie key), *date*, and *search_id*.

Create an endpoint in the API, `/api/v1/search/event/` that allows the UI to perform POST and PUT requests for creating and updating search events. In response to a POST request for a new search, the API will create a new event, assign a *search_id* value, and return the *search_id* in the response object. Subsequent PUT requests to update the existing event with click information will include the *search_id* value.

Whenever the user modifies the search filter, the UI will initiate the creation of a new event by issuing a POST request.

[...]

Issue #1013, opened 2018-08-03
Proposal: API for logging JavaScript Events in Angular


newswangerd:
## Problem

Support for JavaScript event tracking is currently being added in #995. We need to determine how the data from the Galaxy UI will be sent and ingested by the API.

### Data Format
Send the tracking data from the UI to a single API endpoint using the following format:
```
{
  type: 'event_type',
  payload: {
     // Specific attributes for the event type.
  }
}
``` 

Right now we are aiming to support tracking for link and button clicks, which represent two different types and might look like this:

Link data:
```
{
  type: 'link',
  payload: {
    href: "https://www.galaxy.com/davidwittman/redis",
    name: "redis",
    current_page: "/search"
  }
}
``` 

Button data:
```
{
  type: 'button',
  payload: {
    name: "Read Me",
    current_page: "/davidwittman/redis"
  }
}
```

New event types such as page views and errors can be added as needed. 

### Django REST API
The back end would implement an API endpoint at `api/vx/js_event/` that would only accept post requests in the format described above. We would set up the endpoint to triage requests based on the event type, and store the payload data in the database accordingly.

### Securing the API
Exposing a public API for tracking usage metrics is somewhat risky because it opens up the opportunity for people to spam the API to inflate metrics for their content on Galaxy. I would like to do some basic mitigation for this by only accepting requests that have a `User-Agent` HTTP header and a valid CSRF token from Django, while this doesn't really guarantee that people won't spam the API, it adds a few basic roadblocks to prevent that from happening. 

I'm not really sure how to solve this problem and feedback is much appreciated.

### Error Messaging
Since the health of this API endpoint doesn't really affect people's ability to use Galaxy, I would like to set up the endpoint so that it always returns a HTTP 200 return code, regardless of whether or not the data is accepted. Error messages can be surfaced to the dev team for debugging and monitoring via the console. This would also help make it harder for people to spam the endpoint. 

### Sessions for all Users
Right now we only create session variables for users once they log in. It would be nice to be able to attach a session ID to these API calls so that we can group logging data from the same session together. In order to protect people's privacy, the session should not be correlated nor be able to be correlated to any of the user's identifiable attributes (name, username, email etc.) Sessions can be tracked by setting a randomized cookie with a short (15 minute) lifespan that gets renewed each time the browser makes an API request to galaxy.

cutwater:
Can you please describe what problem this proposal is intended to solve? It would be great if you write it in "Problem" paragraph.

[...]

PR #1114, opened 2018-08-29
User session tracking

chouseknecht:
Attempting to bring together #1111 and #1013 in a single solution to be flexible enough to capture any UI events we want to track, but initially focused on search.

Here's what it does so far: 

- Assigns a unique ID to each site visitor, even anonymous (or authenticated) visitors
- Provides `/api/v1/events/` endpoint for creating UI events
- Endpoint accepts *event_type* and *event_data* as params, where *event_data* is a JSONField
- Only accepts a defined list of *event_type* values, which for now only contains 'search'
- Cleans and formats incoming JSONField data for 'search' events

For each user, each unique search is tracked, including query params, number of results returned, and subsequent clicks on next page, prev page, page size, and content item.

The resulting data structure for a search looks like the following:

```
{
            "id": 5,
            "url": "/api/v1/events/5/",
            "related": {},
            "summary_fields": {},
            "modified": "2018-08-30T21:55:01.141920Z",
            "session_id": "52722f71-d52c-456c-9cd4-ae2a29c11536",
            "event_type": "search",
            "event_data": {
                "search_results": 0,
                "search_params": {
                    "keywords": "nginx",
                    "deprecated": "false"
                },
                "next_page_clicks": 3,
                "prev_page_clicks": 2,
                "repositories_clicked": [
                    26290,
                    26763,
                    44081,
                    31119,
                    59668,
                    58767,
                    58741,
                    34933
                ],
                "results_clicked": [
                    12,
                    81,
                    85,
                    92,
                    1058,
                    1045,
                    1043,
                    1
                ],
                "content_items_clicked": [
                    7334,
                    7509,
                    16232,
                    9476,
                    26502,
                    25918,
                    25890,
                    1907
                ]
            },
            "active": null
        },
```


galaxy/api/serializers/event.py
```diff
@@ -0,0 +1,124 @@
...
+    class Meta:
+        model = models.SessionEvent
+        fields = (
+            'id',
+            'url',
+            'related',
+            'summary_fields',
+            'created',
+            'modified',
+            'session_id',
```
newswangerd: I would prefer it if we didn't make user's actual session IDs available to the public via the API. It seems like these could be used to potentially link searches back to individual users, or make it easier for attackers to spoof other people's sessions. 

I propose storing SHA hashes of the session IDs in the database instead. That way we can still aggregate events by a unique session identifier, but we can't look up session IDs that are linked to a specific user's session without knowing the real session ID, which should only be available to the individual user.

galaxy/api/access.py
```diff
@@ -391,6 +391,22 @@ def can_delete(self, obj):
...
+class SessionEventAccess(BaseAccess):
+    model = models.SessionEvent
+
+    def can_read(self, obj):
+        return True
+
+    def can_add(self, data):
+        return True
+
+    def can_change(self, obj, data):
```
newswangerd: maybe make it so that you can only change an object if the session ID's match

galaxy/api/serializers/event.py
cutwater: Completely agree. Considering current implementation of ACL (`SessionEventAccess` class) it's a huge security issue.

galaxy/api/access.py
```diff
@@ -391,6 +391,22 @@ def can_delete(self, obj):
         return False
 
 
+class SessionEventAccess(BaseAccess):
+    model = models.SessionEvent
+
+    def can_read(self, obj):
+        return True
```
cutwater: Do we really want to allow anybody access to our internal session tracking information?
I would say it should be push-only.

galaxy/api/access.py
cutwater: Agree

galaxy/api/serializers/event.py
```diff
@@ -0,0 +1,124 @@
...
+    'EventSerializer',
+)
+
+logger = logging.getLogger(__name__)
+
+
+class ValidSearchType(object):
+    choices = []
+
+    def __init__(self):
```
cutwater: Why do you initialize static class field in the `__init__` method?

```
class ValidSearchType(object):
    choices = [tp.value for tp in constances.EventType]
```

galaxy/api/serializers/event.py
```diff
@@ -0,0 +1,124 @@
...
+        for choice in constants.EventType.choices():
+            self.choices.append(choice[0])
+
+    def __call__(self, value):
+        if value not in self.choices:
+            message = '%s is not a valid choice.' % value
+            raise drf_serializers.ValidationError(message)
```
cutwater: Actually you could replace this whole class with simple function:
```
def valid_search_type(value):
    try:
        constants.EnumType(value)
   except ValueError:
       raise drf_serializers.ValidationError(...)
```

galaxy/api/serializers/event.py
```diff
@@ -0,0 +1,124 @@
...
+        return reverse('api:event_detail', args=(instance.pk,))
+
+    def get_releated(self, instance):
+        return{}
+
+    def get_summary_fields(self, instance):
+        return {}
+
+    def update(self, instance, validated_data):
+        event_data = clean_search_data(copy.copy(instance.event_data))
```
cutwater: Why copy is needed here?

galaxy/api/serializers/event.py
```diff
@@ -0,0 +1,124 @@
...
+        'next_page_clicks': data.get('next_page_clicks', 0),
+        'prev_page_clicks': data.get('prev_page_clicks', 0),
+        'results_clicked': data.get('results_clicked', []),
+        'repositories_clicked': data.get('repositories_clicked', []),
+        'content_items_clicked': data.get('content_items_clicked', []),
+        'last_page_size': data.get('last_page_size', 0)
+    }
+
+
+def append_list_to_list(source_data, target_data, label):
```
cutwater: It's unclear what this function does. Docstring or some explanation is required.
Function name indicates common used util function, but instead it implements some context specific business logic.

galaxy/api/access.py
newswangerd: if we change `SessionEvent` to store session hashes, does it matter if people can read the hashes?

PR #1129, opened 2018-09-04
Adds API endpoint for verifying emails.

newswangerd:
Adds the following endpoints

#### /api/v1/emails/EMAIL_ID/verification/
Accepts: GET, POST
```
POST
{
  'email_address': EMAIL_ID
}
```
If the email address isn't verified, creates an `EmailConfirmation` object for the email address specified in `EMAIL_ID`

GET: returns an empty string.

#### /api/v1/emails/EMAIL_ID/verification/VERIFICATION_KEY
Accepts: GET

Loads the `EmailConfirmation` object for `VERIFICATION_KEY`. If the `EmailVerification` object is the same as the one specified in `EMAIL_ID`, the email is set to verified.


galaxy/api/access.py
```diff
@@ -391,6 +391,30 @@ def can_delete(self, obj):
...
+    # Emails are verified via GET request. This means that emails can be
+    # verified without requiring authentication if a confirmation key exists.
+    def can_read(self, obj):
+        return True
+
+    def can_add(self, data):
+        if self.user.is_authenticated():
+            if data.get('email_address') is None:
+                return True
+            email = EmailAddress.objects.get(pk=data.get('email_address'))
```
chouseknecht: This should be wrapped in a *try/except* that catches *ObjectDoesNotExist*. Otherwise, it's going to fail in with 500 error, I think.

PR #1131, opened 2018-09-05
Added generic services

newswangerd:
Added two new classes: `GenericQuery` and `GenericQuerySave`. The purpose of these classes it to handle the query logic for most of our angular Services. In most cases our services just need to define two functions `query()` and `save()`. Generally the functionality of these two functions is the same across all our services, so they can be abstracted into generic services.

To use the generic queries simply create a service that extends `GenericQuery` or `GenericQuerySave` and pass in a type that the service returns as well as the URL for the API and name of the service like so:

```
export class SurveyService extends GenericQuerySave<Survey> {
    constructor(http: HttpClient, notificationService: NotificationService) {
        super(http, notificationService, '/api/v1/community_surveys', 'survey');
    }
}
```

In this case `Survey` is passed as the type for `GenericQuerySave`, the API URL is set to `/api/v1/community_surveys` and the service name is set to `survey`.

### DIFF
--- a/galaxyui/src/app/resources/survey/survey.service.ts
+++ b/galaxyui/src/app/resources/survey/survey.service.ts
@@ -1,59 +1,14 @@
-import { HttpClient, HttpHeaders } from '@angular/common/http';
+import { HttpClient } from '@angular/common/http';
 import { Injectable } from '@angular/core';
 
-import { Observable, of } from 'rxjs';
-import { catchError, map, tap } from 'rxjs/operators';
-
 import { NotificationService } from 'patternfly-ng/notification/notification-service/notification.service';
-import { PagedResponse } from '../paged-response';
 import { Survey } from './survey';
 
-const httpOptions = {
-    headers: new HttpHeaders({
-        'Content-Type': 'application/json',
-    }),
-};
+import { GenericQuerySave } from '../base/generic-query-save';
+
 @Injectable()
-export class SurveyService {
-    private url = '/api/v1/community_surveys';
-
-    constructor(private http: HttpClient, private notificationService: NotificationService) {}
-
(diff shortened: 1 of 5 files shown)


galaxyui/src/app/resources/survey/survey.service.ts
```diff
@@ -1,59 +1,14 @@
...
-    }),
-};
-@Injectable()
-export class SurveyService {
-    private url = '/api/v1/community_surveys';
-
-    constructor(private http: HttpClient, private notificationService: NotificationService) {}
-
-    query(params?: any): Observable<Survey[]> {
-        params.page_size = 1000;
```
chouseknecht: I don't see a default *page_size* represented in the generic class. Is this going to create a problem? Setting a page size of 1000 was a lazy way around having to handle paging. The default API page size is 10, so if more than 10 items exist, and the page size is not set to a value greater than the total number of items , the query method would need to recursively call the API to retrieve all pages.

chouseknecht: I don't see a default page_size represented in the generic class. Is this going to create a problem? Setting a page size of 1000 was a lazy way around having to handle paging. The default API page size is 10, so if more than 10 items exist, and the page size is not set to a value greater than the total number of items , the query method would need to recursively call the API to retrieve all pages.

newswangerd: It shouldn't cause problems because I moved the `page_size` param here: https://github.com/ansible/galaxy/pull/1131/files#diff-f186fcc519707d031dbbd1eae2d89ba4R93

newswangerd: Is there a better way to load all of the surveys than set the page size to 1000?

PR #1143, opened 2018-09-12
Fix typo

Gisson:
In file examples.rst there was a typo in the word `default` it was instead `deault`


chouseknecht:
@Gisson 

Thanks for catching this!

PR #1150, opened 2018-09-18
Allow hyphens and underscores in tags

redhatrises:
This feature allows hyphens and/or underscores in tags. This is useful for spaces or identifiers e.g. nist_800-53 which is more accurate for a tag than nist80053

### DIFF
--- a/galaxy/constants.py
+++ b/galaxy/constants.py
@@ -24,7 +24,7 @@ import re
 
 MAX_TAGS_COUNT = 20
 PROVIDER_GITHUB = 'GitHub'
-TAG_REGEXP = re.compile('^[a-z0-9]+$')
+TAG_REGEXP = re.compile('^[a-z0-9_-]+$')
 
 
 class Enum(enum.Enum):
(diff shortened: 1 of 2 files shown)


chouseknecht:
@redhatrises 

Don't think we want to do this. We specifically don't want spaces, special chars, nor multiple words in tags.

redhatrises:
Thanks for the response @chouseknecht. Sorry for the delay in responding back.

> Don't think we want to do this. We specifically don't want spaces, special chars, nor multiple words in tags.

Why is this? What it the reasoning for it? I completely agree with the majority of special characters and spaces but not with hyphens and underscores as the tagging is limited and incomplete without them. For our Public Sector customers, we need a way for them to search galaxy_tags for profiles configured as `800-53` or `800-171` which is a win for their usage of galaxy.

shawndwells:
+1 to this RFE. Would greatly enhance readability of tags.

cutwater:
Closing this PR until the decision is made. Please use #1563 to discuss this issue.

PR #1151, opened 2018-09-18
Enable authentication on mobile devices

### LINKED ISSUE #1148: Can't Log Into Galaxy on Mobile
Known issue, and not straightforward to fix. When the screen size is small, and the left nav is replaced by the drop-down menu, the drop-down menu contains no login option.

chouseknecht:
Fixes #1148

### DIFF
--- a/galaxyui/src/app/app.component.ts
+++ b/galaxyui/src/app/app.component.ts
@@ -147,10 +147,11 @@ export class AppComponent implements OnInit {
                     value: me.staff ? 'Staff' : 'User',
                 });
 
-                this.addNavButtons();
+                this.addContentButtons();
             } else {
-                this.removeNavButtons();
+                this.removeContentButtons();
             }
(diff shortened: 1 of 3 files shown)


galaxyui/src/app/app.component.ts
```diff
@@ -201,13 +202,75 @@ export class AppComponent implements OnInit {
...
+        let i: number = null;
+        this.navItems.forEach((item: VerticalNavigationItem, index: number) => {
+            if (item['title'] === title) {
+                i = index;
+            }
+        });
+        return i;
+    }
+
+    private optionallyAddLoginButtons(): void {
```
newswangerd: I think you can set `mobileItem` to hide items when PF is not in mobile mode.

```
             {
                 title: 'My Imports',
                 iconStyleClass: 'fa fa-upload',
                 url: '/my-imports',
                 mobileItem: true,
            },
```

Issue #1168, opened 2018-09-21
Content scoring improvements


awcrosby:
Continue improvements on the past two content scoring iterations:

- [ ] For the importer checks that emit warnings/errors, track these as a new linter type violation so they can impact the quality score
- [ ] Add two rule_id: severity mappings to match yamllint codes of ‘error’ and ‘warning’
- [ ] Disable flake8 linter for now until better strategy on how it should impact quality score
- [ ] Move rule_id: severity mapping from database into code/config

PR #1174, opened 2018-09-26
Content scoring improvements

awcrosby:
Works #1168


galaxy/worker/importers/role.py
```diff
@@ -162,14 +182,18 @@ def _add_dependencies(self, role, dependencies):
                     role.dependencies.add(dep_role)
                     new_deps.append(dep)
                 except Exception as e:
-                    self.log.error(
-                        u"Error loading dependencies {}".format(e))
+                    msg = u'Error loading dependencies {}'.format(e)
+                    self.linter_data['linter_rule_id'] = 'dependency_load'
+                    self.linter_data['rule_desc'] = msg
+                    self.log.warning(msg, extra=self.linter_data)
```
awcrosby: This was changed from error to warning since it doesn’t stop the importer from importing

galaxy/main/migrations/0118_scoring_edits_remove_contentrule.py
```diff
@@ -0,0 +1,33 @@
...
+
+    dependencies = [
+        ('main', '0117_namespace_search_index'),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
+            name='ContentRule',
+        ),
+        migrations.AddField(
```
awcrosby: This new migration removes the ContentRule object added in an earlier migration

galaxy/main/migrations/0118_scoring_edits_remove_contentrule.py
```diff
@@ -0,0 +1,33 @@
...
+# Generated by Django 1.11.15 on 2018-09-25 19:35
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0117_namespace_search_index'),
```
newswangerd: I added a migration named `0118_scoring_edits_remove_contentrule` in #1176 that has the same dependency as this. You'll need to rebase against devel and update this to reference the migration that I just added

Issue #1056, opened 2018-08-10
Enabling users to mark the content collections they are interested in following


tima:
### Use Case

As a Galaxy content consumer I would like to mark collections I am interested in or using so I can follow changes and enhancements over time more easily.

### Proposed Solution

File this under providing more features and values to consumers of content thru Galaxy.

Similar to Github, users should be able follow (star?) a collection that they are using or interested in. This information could then be utilized in numerous ways to make finding and tracking content in Galaxy. Just to name a few ideas:

* Quick navigation to the content they care about on their My Content page.
* Displaying activity on the content they are following -- new releases, changes in status etc.
* Displaying other users interest and [trending content](https://medium.com/hacking-and-gonzo/how-hacker-news-ranking-algorithm-works-1d9b0cf2c08d)
* Prompting for feedback aka [community surveys](https://github.com/ansible/galaxy/issues/948)

The only drawback is that user would need a Github account and sign into Galaxy. This feature would not be anonymous. 

### Alternatives

None come to mind. The base idea is flexible enough that it could be taken numerous ways.

### Implementation

The proposed idea covers the specifics for the base feature. Look no farther then the right hand corner of the page this issue is on. Something like a star button and a count. Things like activity feeds, trending collection lists etc. could follow separately.

PR #1195, opened 2018-10-11
Add the ability to follow users and collections

newswangerd:
#1056

screen shot 2018-10-11 at 9 51 45 am
screen shot 2018-10-11 at 9 51 57 am


galaxy/accounts/migrations/0008_auto_20181011_0950.py
```diff
@@ -0,0 +1,26 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name='customuser',
+            name='repositories_followed',
+            field=models.ManyToManyField(blank=True, related_name='following_users', to='main.Repository'),
+        ),
+        migrations.AddField(
+            model_name='customuser',
+            name='users_followed',
```
chouseknecht: Should this be 'namespaces_followed'? You don't follow a user, you follow a Namespace, even though a username and Namespace name might be the same, the objects are different.

chouseknecht:
@tima @newswangerd 

We need to start decoupling from GitHub. Should we remove GitHub Watchers, and replace with Galaxy Followers?

cutwater:
Please don't forget to squash you commits and double check commit message before merge.

PR #1198, opened 2018-10-12
Add explicit `on_delete` parameter for `ForeignKey`

cutwater:
Since Django 2.0 `on_delete` parameter of `ForeignKey.__init__`
has become required. This patch adds missing on_delete parameters
equal to former default value `CASCADE`.


cutwater:
@newswangerd I can't find any `ForeignKey` fields in `galaxy/accounts/models.py`. If you're referring to `ManyToManyField` fields, they're not affected by this change.

PR #1239, opened 2018-10-24
Fix default value for JSON fields

cutwater:
Mutable types must not be used as a default value for
JSONField instances as it will cause multiple
model objects having a references to a shared object.
A callable should be used instead.

See https://docs.djangoproject.com/en/2.1/ref/contrib/postgres/fields/#django.contrib.postgres.fields.JSONField


newswangerd:
Not as important, but would add a few commends explaining why we're overriding `__init__()` and `from_db()` as well as not just initializing `preferences` from `DEFAULT_PREFERENCES`? I think that will help people in the future to not make the same mistake I did here.

Issue #1234, opened 2018-10-23
Docs: develop guidelines for documenting certified roles


acozine:
## Feature Request
Documentation guidelines for certified roles.

### Use Case
As a partner, I want clear requirements for getting certification for my roles, including details about what documentation I must include, in what format, etc.

### Proposed Solution
- [ ] review current network roles and note what works and what doesn't work in the current role documentation
- [ ] use findings to create guidelines for the location, required content, and format of role documentation

acozine:
See https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#module-documenting for module guidelines, for comparison.

acozine:
Related to https://github.com/ansible/proposals/issues/19.

Issue #1236, opened 2018-10-23
Docs: Try porting one role's docs from markdown to YAML or rST


acozine:
## Feature Request
Consider options for publishing role documentation as Galaxy continues to evolve. One option is to publish role docs in the same way we publish module docs. 

### Use Case
As a technical writer, I want to explore the best way to publish role documentation for certified content. 
As a partner or role developer, I want clear guidelines on the format required for my documentation.

### Proposed Solution
Test out porting the documentation on one role from markdown to YAML (to be transformed to rST as we do with module docs) or to reStructuredText. Is this reasonable to expect of partners? Does it provide advantages over the current system of publishing a single README.md for each role?

geerlingguy:
Just wondering—in a lot of these discussions around rST / YAML / etc. would the intention be to drop support for markdown, or add additional support for other formats?

Because as someone who loves the fact that markdown 'just works' almost everywhere nowadays (it's kind of the 'plain text' of structured text formats), it would be disappointing to have to maintain docs in a different format just for Galaxy content.

acozine:
@geerlingguy that's great feedback, thanks. We haven't formulated a plan yet - at this point we're gathering information and testing several options, including using `recommonmark` to process markdown files with Sphinx.

ericsysmin:
@chouseknecht any update to this?

PR #1269, opened 2018-10-31
add documentation guidelines for Galaxy roles

samccann:
Fixes https://github.com/ansible/galaxy/issues/1234

Add a section to the Galaxy contributor docs to cover how to document a role, with some examples.

### DIFF
--- a/docs/docsite/rst/contributing/documenting_content.rst
+++ b/docs/docsite/rst/contributing/documenting_content.rst
@@ -4,7 +4,7 @@
 Documenting Galaxy content
 **************************
 
-.. contents:: Topics
+.. contents::
    :local:
 
 This topic describes how to document Ansible content that can be imported into Galaxy.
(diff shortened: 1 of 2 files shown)


chouseknecht:
@samccann 

Not opposed to this. What's the bigger picture though? What do we expect the Galaxy server to do with `docs/*.md`? Do we expect `ansible-doc` to read these files?

samccann:
The attempt here is to put in some guidelines so that we know roles (especially certified roles, as they come in) will have good documentation to go with it.  Longer-term, we need to figure out the strategy as you are suggesting. Personally, yes, I'd like to see `ansible-docs` work for roles as well. I don't know at the moment what that entails, but as roles grow, we want as much documentation support as we have today for modules.  

There is a longer term need to figure out what Galaxy does with those `/md `files. There's another issue related to that which is more of an experiment - [](https://github.com/ansible/galaxy/issues/1236). 
..adding @acozine as well. We are just starting out with thoughts on how Galaxy documentation/information needs to evolve as roles use grows.

thaumos:
Is it possible to have a call about this, @samccann?

I also am interested in having ansible-doc handle collection documentation as well. It's not something on the roadmap right now for Engine, but we can talk about it for later next year. 

I want to understand the choices in doc format type out of curiosity.  Also just hear any other thoughts you all may have on the subject.

samccann:
Sent something to your calendars for later next week. thanks!

cutwater:
@chouseknecht @samccann what is the status of this PR?

samccann:
@cutwater on hold. We just restarted the discussion around what is required for roles documentation. I'm closing this PR for now.

PR #1274, opened 2018-11-02
Remove old search metrics collection system.

newswangerd:
Removes support for tracking search queries using prometheus (added in #989) because we are inserting metrics directly into influx rather than using Prometheus.


galaxy/settings/development.py
```diff
@@ -92,9 +92,3 @@
 ]
 
 STATIC_ROOT = ''
-
-# =========================================================
-# Metrics Settings
-# =========================================================
-
-METRICS_ENABLED = True
```
cutwater: I think we should keep `METRICS_ENABLED` setting and adopt it for new metrics implementation, so that InfluxDB service can be optional for development environment.

newswangerd: Done. I renamed it to `GALAXY_METRICS_ENABLED`.

PR #1287, opened 2018-11-05
Prevent users from scoring their own content.

newswangerd:
If a user is one of the owners of a namespace, prevent them from scoring any content in that namespace:

screen shot 2018-11-05 at 4 58 04 pm


galaxy/api/serializers/survey.py
```diff
@@ -16,6 +16,8 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 from . import serializers
+from rest_framework import serializers as drf_serializers
```
cutwater: Please arrange imports according to PEP8

Issue #1281, opened 2018-11-03
Unable to import role via Travis hook


otakup0pe:
##### SUMMARY
While I am able to import a role manually via the website, the notification hook doesn not seem to be working properly. The repository I have been testing is `ANXS/apt`. This wasn't working prior to the move from the Travis CI integration to the app either.

##### STEPS TO REPRODUCE

Followed the [guide](https://docs.ansible.com/ansible/2.7/reference_appendices/galaxy.html#travis-integrations). Logged into Galaxy with a fresh GitHub token with the `repo:public` and `user:email` scopes. The scopes were derived from [this](https://github.com/ansible/galaxy-issues/issues/202#issuecomment-254701562) issue comment, as it wasn't clear from the docs. Ran setup on the repository (with both lowercase and uppercase `ANXS` just in case!) and the behavior remained the same. 
I also tried logging out/in to the web interface "just in case".

##### EXPECTED RESULTS

Imports work from both the web ui and the notifications hook.

##### ACTUAL RESULTS

Able to import via the web site, and local commandline, but unable via the hook.

```
failed: 401 {u'documentation_url': u'https://developer.github.com/v3', u'message': u'Bad credentials'}
```

chouseknecht:
@otakup0pe 

This will be released to production tomorrow (7 Nov 2018).

otakup0pe:
@chouseknecht any update on this? Is there something different I should be trying?

PR #1296, opened 2018-11-06
Fix NotificationList api returning exception object

cutwater:


chouseknecht:
@cutwater 

As discussed in chat, can we make the lookup on `provider_namespae` case insensitive. I think that might fix #1281

chouseknecht:
@cutwater 

LGTM. Thanks for fixing this!

otakup0pe:
Thanks y'all I will test it out as soon as I can 💖 Is it on the QA cluster or live?

Issue #914, opened 2018-07-17
Re-import not finding latest release


trevorr:
##### SUMMARY

I'm attempting to re-import an existing role to pick up the latest GitHub release. The import seems to completely successfully, but `ansible-galaxy install -f -r ansible/requirements.yml` still finds the previous version. The requirements file lists no version:

```
- src: trevorr.consul
```

##### STEPS TO REPRODUCE

Go to https://galaxy.ansible.com/my-imports?namespace=trevorr&repository_name=consul&selected=263305&page_size=10 and click the re-import button. Here is the import output:

```
INFO Starting import: task_id=263305, repository=trevorr/ansible-consul
INFO Content search - Looking for file "apb.yml"
INFO Content search - Looking for top level role metadata file
INFO [role] Linting...
ERROR [role] ./tasks/install-cli.yml:1:1: [error] too many blank lines (1 > 0) (empty-lines)
ERROR Lint failed
INFO [role] Found Content instance: id=11574, content_type="role", name="consul"
INFO [role] Adding role metadata tags
INFO [role] Adding role platforms
INFO [role] Adding role dependencies
INFO Updating repository versions...
Import completed
```

I'm assuming that the lint error is not fatal, since #824 has been fixed. I'm not sure why the lint error occurs either, since it doesn't happen when I run `ansible-lint` locally. `empty-lines` in the Galaxy importer appears to be set to `max: 3`. I could delete the leading blank line, but this is a fork that I don't want to change gratuitously.

##### EXPECTED RESULTS

I expected the [latest release](https://github.com/trevorr/ansible-consul/releases/latest) (1.9-tr3) to be downloaded.

I would also expect to see the GitHub release versions/tags to be displayed, as they used to be in the Galaxy UI, but I don't see them anywhere. The [displayed version](https://galaxy.ansible.com/trevorr/consul) is 1.0.0, which isn't listed anywhere in the role. Perhaps it's a default value?

##### ACTUAL RESULTS

The previous release (1.9-tr2) is downloaded.

```
ansible-galaxy install -f -r ansible/requirements.yml
- downloading role 'consul', owned by trevorr
- downloading role from https://github.com/trevorr/ansible-consul/archive/1.9-tr2.tar.gz
- extracting trevorr.consul to /Users/trevor/ansible-roles/trevorr.consul
- trevorr.consul (1.9-tr2) was installed successfully
```

##### NOTES

If I explicitly include the desired GitHub release version:

```
- src: trevorr.consul
[...]

PR #1321, opened 2018-11-08
Add importer warning when git tag is not in semver format

awcrosby:
Closes: #914 

Adds a single importer warning listing all git tags not semver compatible, example:
`Galaxy will only import git tags that match the semantic version format, skipping these tag(s): v3.0, v3.0-beta, v2.2, v2.0, v2.1, v2.0-beta, v1.0, 3.0`


brtknr:
This is awesome! Thank you :)

PR #1326, opened 2018-11-09
Make explicit about how to now do a Local Install

thadguidry:
Helpful folks on the mailing list, @sivel was able to point me to the contributing guide as a reference to installing locally.  This should have been mentioned in the README.md and a few other notes about the old docker image that the old installer used and is now considered deprecated since 3.0

I've tried to make it super clear about what is going on with these changes.  Corrections welcome if I overspoke anywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,9 @@ For help using the public Galaxy web site to find and install Ansible content, o
 
 ## Installing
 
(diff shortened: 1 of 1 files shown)


chouseknecht:
@thadguidry Thank you for the updates to our README. Apologies for not merging this sooner.

PR #1335, opened 2018-11-12
Update search ranking algorithm.

newswangerd:
This updates the best match formula to:

screen shot 2018-11-12 at 4 01 55 pm

Where: 
k = keyword rank (0 to 1)
c = community score (0.005 to 0.015)
d = number of downloads
q = quality score (0 to 5)

Explanation:
This formula is broken into three components:  keyword, community/download score and quality score.

**Keyword**
Value received from postgres

**Community/download**
The community score is used as a modifier on the download count. This essentially allows for the weight of the download count to be lowered by lowering the community score.  Since the download count is factored in logarithmically this makes it so that large download counts will be relatively unaffected by community scores while content with low download counts will be dramatically affected by community scores.

This graph shows the result of this function on content that has received a community score of 5 (which corresponds to a modifier of 0.015) and a score of 0 (which corresponds to 0.005). Keeping the maximum modifier bellow 1 makes it so that the scale increases a lot more gradually in the beginning, allowing for more granularity at lower numbers of downloads.

screen shot 2018-11-12 at 3 39 17 pm

In comparison, this is what the current function looks like. As you can see, the boost from the quality score is far too pronounced and the graph ramps up far too quickly initially.

screen shot 2018-11-12 at 3 47 04 pm

**Quality**
This allocates between 0 and 0.2 points based on the quality score. We expect that the quality scores will mostly be 0 or 5 since people will either not bother re importing their content, or they’ll put in a little effort to make their content get a 5. This uses a logarithmic scale so that a small decrease from 5 to 4.5 or 4 points has a relatively low effect on the search ranking. The rationale behind this is that there are inevitably going to be cases where people have to violate linter rules to make things work (such as using shell or command modules to call git in cases where the git module can't be used). This should mean that users with a small number of inevitable violations perform better than users with larger numbers of violations.

This portion of the formula will have a fairly large impact on best match regardless of the number of downloads a collection has. This is fine because we want good incentives to get people to improve their quality scores.

### DIFF
--- a/galaxy/api/views/search.py
+++ b/galaxy/api/views/search.py
@@ -51,6 +51,8 @@ RANK_NORMALIZATION = 32
 
 DOWNLOAD_RANK_MULTIPLIER = 0.4
 CONTENT_SCORE_MULTIPLIER = 0.2
+COMMUNITY_SCORE_MODIFIER = 0.002
+COMMUNITY_SCORE_MODIFIER_MIN = 0.005
 
 
 class ApiV1SearchView(base.APIView):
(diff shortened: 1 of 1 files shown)


cutwater:
@newswangerd Thank you for working on this issue. Looks good to me.

newswangerd:
@awcrosby that's a really good point. I'm going to have to think about that some more.

alikins:
Are there tests that check some known repositories search results to verify they show up as expected?

newswangerd:
@alikins nope. Just manual tests.

Issue #1352, opened 2018-11-17
Travis badge not working for roles imported from travis-ci.com


Bassinator:
##### SUMMARY
For all my roles imported from travis-ci.com the build status is shown unknown.

The link of the badge is pointing to travis-ci.org. I guess that's the reason for this issue.

##### STEPS TO REPRODUCE
Import ansible role and notify from travis-ci.com

##### EXPECTED RESULTS
correct badge with actual build result is shown

##### ACTUAL RESULTS
build status unknown

```

```

chouseknecht:
@newswangerd 

I suspect we're hardcoding the badge link in the API to `travis-ci.org`.

Bassinator:
If you want to keep these hard coded. I would suggest to stick it to travis-ci.com because all builds on travis-ci.org will be migrated to travis-ci.org.

newswangerd:
@Bassinator the fix for this problem should be going into production sometime tomorrow.

Bassinator:
I can confirm, that this is fixe. Build state is indicated with the correct badge after the next run on travis.

PR #1366, opened 2018-11-26
Derive travis badge from travis build URL.

newswangerd:
Fixes travis build status when launching imports from travis-ci.com.

#1352


newswangerd:
Note, this needs to be tested in galaxy-dev because there's no way to get travis to send the test info to localhost.

PR #1368, opened 2018-11-27
Remove old import task msg on import, and managment command

### LINKED ISSUE #1349: Import Cleanup
During the 3.1 deployment the migration for import_task_messages failed. It ran into a deadlock while attempting to add new fields. While investigating we discovered the table contains > 10M rows, which is unmanageable at best, and the data is probably not very useful.

We need to define and create a periodic task that cleans this up, or maybe delete old imports when a new one happens for a given repository.

awcrosby:
Deletes the import task messages for the previous import. Works #1349 

This PR also includes a django managment command (for one-time manual use) that for each repo removes all import task messages except those tied to the latest import.


galaxy/main/management/commands/remove_old_import_task_msg.py
```diff
@@ -0,0 +1,28 @@
+from django.core.management.base import BaseCommand
```
cutwater: Minor: Naming, maybe better to call it `import_task_message_cleanup`

galaxy/worker/tasks.py
```diff
@@ -349,6 +350,20 @@ def _update_quality_score(import_task):
     LOG.debug(u'repo quality score: {}'.format(repository.quality_score))
 
 
+def _remove_last_task_msg(import_task):
+    last_task = models.ImportTask.objects.filter(
```
cutwater: Probably it makes more sense not limiting cleanup by only last import task and just cleanup all import tasks excluding current task.

PR #1406, opened 2018-12-07
Weight community score based on number of surveys submitted

### LINKED ISSUE #1401: Role community score dramatically affects overall score, often negatively
##### SUMMARY

I was glancing at a few of my roles on Galaxy as I was doing some manual cleanup work, and noticed that one role in particular (`geerlingguy.java`) suddenly had an overall rating of 2.5/5.

Now, I don't like Java as much as the next person, but I've spent dozens of hours on that role, making sure it works across a huge variety of OSes and versions, and works with a lot of other major Java projects downstream... and was surprised to see such a poor rating!

Anyways, it seems that there has been only one community review so far... and that review was just "0, 0, 0, 0, 0" for everything. I can imagine I could quickly put the kibosh on any new role that is submitted to Galaxy by going in and rating it as 0's across the board, thus dropping it's overall score to at most 2.5/5, then making it much less likely someone would consider trying it out.

##### STEPS TO REPRODUCE

1. Go to any role on Galaxy with 0 community ratings.
1. Rate it as '0' and 'N' for quality scores.

##### EXPECTED RESULTS

This malicious rating should not dramatically impact the chances this role would be evaluated by those using Ansible Galaxy.

##### ACTUAL RESULTS

This single malicious rating drops the role from a possible 5.0 rating (assuming the linting passes muster) to a maximum 2.5 rating, turning the rating value _orange_, which visually indicates the role is not that good (to a user evaluating roles on Galaxy).

newswangerd:
Fixes #1401

Rather than use an unweighted average for the combined score, use a weighted average that's proportional to the number of surveys that have been submitted:

screen shot 2018-12-07 at 10 55 23 am
Where 
- q = quality score
- c = community score
- x = the number of surveys submitted (up to 3)

This would result in the following for a role that has a quality score of 5 and community score of 0:
- 0 surveys -> 5
- 1 survey -> 4.17
- 2 surveys -> 3.34
- 3+ surveys -> 2.5


galaxy/main/models.py
```diff
@@ -1060,6 +1060,12 @@ def content_counts(self):
             .annotate(count=models.Count('content_type__name')) \
             .order_by('content_type__name')
 
+    @property
+    def community_survey_count(self):
```
cutwater: This property will produce +1 query for each repository item in list.
Please consider using join + count.

Issue #1420, opened 2018-12-14
Repository names are not sanitized by removerole API endpoint


ironfroggy:
##### SUMMARY
Importing sanitizes the repository name, which includes replacing dashes `-` with underscores `_`, but this sanitizing is not done when deleting a repository via the CLI/API. Consequently, after a successful import a subsequent attempt to delete with exactly the same parameters will fail_only if the repository name needs to be sanitized.

##### STEPS TO REPRODUCE
- Create a repository on Github with a name containing uppercase letters, dashes, or periods.
- Import this repository, even if it does not contain any importable content.
- Attempt to delete the repository and its imported content, if any, from Galaxy.

##### EXPECTED RESULTS
Any imported content from the repository is removed and the repository itself is removed from your account.

##### ACTUAL RESULTS
The repository and any content from it is not found, because its real name does not match the name it was imported under and the delete path does not sanitize the input name.

`Role ironfroggy.django-better-cache not found. Maybe it was deleted previously?`

PR #1423, opened 2018-12-14
Fix repositories deletion failure when using ansible-galaxy client

cutwater:
ansible-galaxy client uses deprecated API to delete user repositories.
This patch fixes reporisotry and roles lookup performed by `name`
attribute instead of `original_name`.

Issue: #1420


ironfroggy:
Tested locally, looks good! 👍

Issue #1480, opened 2019-01-24
Community score algorithm is broken in some cases?


geerlingguy:
##### SUMMARY

I'm looking around at some of my roles' community scores, and I noticed one weird situation with my https://galaxy.ansible.com/geerlingguy/java role—it has a community score of `3.3/5`, even though the individual components of that score total `23.3/25` (or `4.6/5`).

See screenshot:

<img width="1427" alt="java-role-community-score-weird" src="https://user-images.githubusercontent.com/481677/51715344-f048ca80-1ffd-11e9-8177-4bfef716f44d.png">

##### STEPS TO REPRODUCE

Look at https://galaxy.ansible.com/geerlingguy/java

##### EXPECTED RESULTS

The community score should be `4.6/5` (or something different than the `min` of all the components of the community score).

##### ACTUAL RESULTS

The community score is `3.3/5`.

newswangerd:
This is actually the expected behavior, but you have stumbled on something that I've been thinking about changing because I think it's a little counter intuitive.

The community score is actually an average of the surveys submitted, not the questions. Each survey is the set of 5 questions. In this case you three surveys: one survey with a score of 0 where the guy only answered one question (used in production) and two with a score of 5. This means that the overall score is (0 + 5 + 5) / 15 = 3.333/5.

ironfroggy:
@newswangerd does this mean omitting questions are essentially counted as 0, dragging the average down rather than somehow being likewise omitted in the actual scoring?

Is that the specific aspect you refer to as expected but counter intuitive?

newswangerd:
@ironfroggy Omitted questions are simply not counted. We basically average the average of the surveys. In this case the survey answers look like: {0}, {5, 5, 5, 5, 5}, {5, 5, 5, 5 ,5}. The first one averages 0 and the next two average 5. If the guy from the first survey had answered 5 instead of 0, then the surveys would look like: {5}, {5, 5, 5, 5, 5}, {5, 5, 5, 5, 5} and the average of the averages would be (5+5+5)/15 = 5/5

ironfroggy:
@newswangerd right, what I'm getting at is the effect of this is we're counting survey results of {0} as if they were {0, 0, 0, 0, 0} and I'm asking if that's an intentional or accidental equivalency.

[...]

PR #1486, opened 2019-01-28
Set community score to average of the questions.

newswangerd:
fixes: #1480


galaxy/main/migrations/0126_community_score_question_average.py
```diff
@@ -0,0 +1,31 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0125_collection_base'),
+    ]
+
+    operations = [
+        migrations.RunSQL(SET_COMMUNITY_SCORE_TO_QUESTION_AVERAGE),
+    ]
```
ironfroggy: If `update_community_score()` where part of the Model or otherwise easily available outside the view itself, we could update the existing repository scores using the same code path you've already updated in this PR, ensuring no chance of a mismatch between two different implementations of the algorithm in different languages, one of which exists only in a migration and will only be run once. (Notoriously difficult to validate!)

This would also be a safer setup for future changes to the scoring algorithm, which has already changed a few times and surely will continue to do so.

cutwater: Thoughts:
1. Performing data migrations on can be bottleneck and affect service availability.
2. Running python will affect performance drastically, because instead of single SQL we will end up with sending data back and forth within the single transaction.
3. Re-using code in migration is dangerous, because code can eventually change that WILL break the migration.
4. The best option in this particular case I think to create a background task to recalculate community scores and run it after migration.

newswangerd:
Before we merge this, I want to add some info to the documentation about the community score.

Issue #1522, opened 2019-02-05
500 Server Error from /api/v1/emails/


ironfroggy:
##### SUMMARY
The `emails` endpoint of the API erroneously handles requests without authentication with an internal error.

##### STEPS TO REPRODUCE
```$ curl https://galaxy.ansible.com/api/v1/emails/
<h1>Server Error (500)</h1>
```

##### EXPECTED RESULTS
A valid JSON response that is either an empty result set or a valid error message.

##### ACTUAL RESULTS
A 500 error page in a non-JSON response a client could not parse.

```
<h1>Server Error (500)</h1>
```

PR #1548, opened 2019-02-12
Raise forbidden error if un authenticated users load email api.

newswangerd:
fixes:  #1522

### DIFF
--- a/galaxy/api/views/email.py
+++ b/galaxy/api/views/email.py
@@ -12,6 +12,8 @@ from galaxy.main.celerytasks import user_notifications
 
 from rest_framework.response import Response
 
+from rest_framework.permissions import IsAuthenticated
+
 __all__ = [
     'UserEmailList',
     'EmailList',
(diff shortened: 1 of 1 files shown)


galaxy/api/views/email.py
```diff
@@ -24,13 +24,20 @@
...
+
+
 class UserEmailList(base_views.SubListAPIView):
     model = EmailAddress
     serializer_class = serializers.EmailSerializer
     parent_model = User
     relationship = 'emailaddress_set'
 
     def get_queryset(self):
+        deny_anonymous_users(self.request.user)
```
chouseknecht: I *think* this should be handled in [access.py](https://github.com/ansible/galaxy/pull/1548)

cutwater: Definitely `get_queryset()` is not the best place for checking user permissions.
Please use     `permission_classes = (IsAuthenticated,)`

PR #1572, opened 2019-02-17
Add flake8 GitHub Action to the repo workflow

webknjaz:


.github/main.workflow
```diff
@@ -0,0 +1,8 @@
+workflow "Code linting" {
+  on = "push"
+  resolves = ["GitHub Action for Flake8"]
+}
+
+action "GitHub Action for Flake8" {
+  uses = "cclauss/GitHub-Action-for-Flake8@0.0.1"
```
cutwater: Isn't it required to pass `args` parameter to the workflow file?

webknjaz: Depending on whether you want to override the default one. But probably you're right..

webknjaz: Done

webknjaz:
@cutwater what are the flake8 settings you use?

cutwater:
@webknjaz we use almost defaults settings, but it seems some extra legacy scripts was checked.
We run `flake8` against `galaxy/` directory.

webknjaz:
@cutwater 
Okay, it should be good to go now!

cutwater:
@webknjaz Looks good. Thanks.

PR #1582, opened 2019-02-20
Add the ansible-lint GitHub Action

webknjaz:
Trying out more fancy stuff now


cutwater:
It's unclear for me what is the benefit of running `ansible-lint` on each build against single testing role that is likely to be replaced \ removed in future.

PR #1586, opened 2019-02-26
Collection data model update

cutwater:


galaxy/main/models/collection.py
```diff
@@ -15,50 +15,85 @@
...
+from django.contrib.postgres import fields as psql_fields
+from django.contrib.postgres import search as psql_search
 from django.db import models
 from pulpcore.app import models as pulp_models
 
+from .namespace import Namespace
 
-class Collection(pulp_models.Content):
+
+class Collection(models.Model):
```
awcrosby: Will this eventually use something derived from our BaseModel?

cutwater: Nope, BaseModel and its base classes hierarchy is a legacy garbage and should be avoided.
However it's good point, that `created` and `updated` fields are nice to have

PR #1587, opened 2019-02-26
Add UI Wizard for Uploading Collections

newswangerd:
### Changes
- Adds the option to upload a collection after clicking on the "Add Content" button under "My Content"
- Refactored the existing repo import as a react container

#### Pick an import method
screen shot 2019-02-26 at 10 46 08 am

#### Select a file
screen shot 2019-02-26 at 10 46 36 am

#### Upload in progress
screen shot 2019-02-26 at 10 59 02 am

#### Upload failed
screen shot 2019-02-26 at 11 00 33 am

#### GitHub import view
screen shot 2019-02-26 at 11 23 30 am


awcrosby:
1. Is it worth stating something like “Import Role from Github” since we will only support single role?
2. Regarding “Upload New Collection” and “Collection already exists” - a collection does not have to be new, just a different version.

ironfroggy:
Both the first two-button prompt and then the file-upload prompt could use some help text about the options and expectations, and likely need a link to documentation about the difference between the options and where the collection to upload comes from.

newswangerd:
@awcrosby I changed the text on the github import button. I think this particular workflow is for creating a new collection, so the "Upload New Collection" makes sense for me. Once we have better support for collections on the my content page I want to make it so that the "Import" button pops up a modified version of this modal that asks for a new version of an existing collection.

@ironfroggy I added some help text to the buttons. I'll add documentation once it's been written (#1591)

awcrosby:
Ah, so since this is for a new collection use case... FYI currently the user will only get the "Collection already exists" in the upload box if they are uploading the same artifact (same sha hash). A new version will be a successful import.

ironfroggy:
I like the succinct text added

newswangerd:
@awcrosby I think that the validation for that has yet to be added on the API. RIght now it just takes whatever error message is returned by the API and displays it (aside from the MIME type, which is checked on the front end).

PR #1593, opened 2019-02-27
Add content name importing log msg for repos and collections

### LINKED ISSUE #1590: Output of import process lost important information
##### SUMMARY
Integration tests started failing yesterday and the problem was tracked down to a change in the output of the import process. Specifically, the output logging changed from messages of `===== IMPORTING ROLE: FOOBAR =====` to `===== LOADING ROLE =====`, breaking tests that depend on this output to monitor and confirm the results of the import and removing the information about what was actually imported entirely (the name isn't even included anymore).

This also *retroactively* changes every previously released version of `ansible-galaxy`, which pipes the task output to the user. Because of this broad retroactive change I'm filing this as a bug because I don't think that scope was considered and, even if this change is deemed necessary, the scope of the change needs a quick conversation to decide if that's the change or if this is a regression that needs to change.

##### STEPS TO REPRODUCE
On `devel` branch import any content with `ansible-galaxy` and look at the output or the import log in the Galaxy UI.

##### EXPECTED RESULTS
Log output including lines like `===== IMPORTING ROLE: FOOBAR =====` indicating imported roles.

##### ACTUAL RESULTS
Log output including lines like `===== LOADING ROLE =====` indicating something was loaded, but not indicating what, and changing the format breaking user expectations and any processes parsing the output.

awcrosby:
Resolves #1590 

Signed-off-by: Andrew Crosby <acrosby@redhat.com>

### DIFF
--- a/galaxy/importer/collection.py
+++ b/galaxy/importer/collection.py
@@ -144,6 +144,10 @@ class CollectionLoader(object):
             loader.lint()
             # content.scores = loader.score()
             self.log.info(' ')
+            self.log.info('===== IMPORTING {}{} ====='.format(
+                          content_type.name, name))
+            self.log.info(' ')
+
             yield content
 
     def _get_collection_quality_score(self):
--- a/galaxy/worker/importers/base.py
+++ b/galaxy/worker/importers/base.py
@@ -45,6 +45,7 @@ class ContentImporter(object):
         content.clean()
         content.save()
 
+        self.log.info(' ')
         return content
 
     def make_content(self):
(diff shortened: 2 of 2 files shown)


galaxy/importer/collection.py
```diff
@@ -144,6 +144,10 @@ def _load_contents(self, content_list):
             loader.lint()
             # content.scores = loader.score()
             self.log.info(' ')
+            self.log.info('===== IMPORTING {}{} ====='.format(
```
ironfroggy: missing a `": "` here between the content type and name, to restore the original behavior

awcrosby: The colon is already in the variable `name`

ironfroggy: 👍

Issue #1528, opened 2019-02-06
Implement collections support in importer 


cutwater:
Implement support for collections along with old-style repositories in galaxy importer.

PR #1599, opened 2019-02-28
Validate collection manifest metadata

awcrosby:
Works #1528


galaxy/importer/utils/spdx_licenses.json
```diff
@@ -0,0 +1,1148 @@
+{
```
webknjaz: @awcrosby where did you take this json from?

awcrosby: it comes from https://github.com/spdx/license-list-data/, using mazer to grab: https://github.com/ansible/mazer/blob/f55f14a64ea738082914a69ca47b9558915ea6ae/Makefile#L67

webknjaz: Ah, I see, thanks :)

PR #1634, opened 2019-03-09
Collection import uses import task and import task messages

awcrosby:
- Write collection import task messages for collection import status via api, and collection scoring
- Improve logging and exception handling
- Edit ImportTask serializer to allow `api/v1/imports` endpoint to display collection import status and messages (likely further edits needed with forthcoming api/v2)


newswangerd:
Right now the collections API endpoint returns something like `{"task":"/api/v1/tasks/3/"}` with the pulp task. It would be great to add the import task id to the response as well so I can redirect users to the import console after they import artifacts from the UI

awcrosby:
> I'm getting this error from the pulp worker when I try to import collections. It looks like it's grumbling about not having a repository associated with the import.
@newswangerd I had to restart the pulp workers and it fixed that

newswangerd:
Ah, that did the trick.

awcrosby:
> It would be great to add the import task id to the response as well

@newswangerd would this also be sufficient?  inside this endpoint: `/api/v1/tasks/3/` was an `import_task_id` or `import_task_url` field, or a summary field with the relevant import_task data and import_task_messages data

newswangerd:
@awcrosby I would much prefer it if it was in the response from the `collections/` api endpoint, if possible. If `import_task_id` is in the `tasks/` endpoint then I have to make an additional query before redirecting the user to the import log.

galaxy/api/serializers/serializers.py
```diff
@@ -346,6 +348,15 @@ def get_related(self, obj):
...
             return {}
         res = super(ImportTaskSerializer, self).get_related(obj)
+
+        if getattr(obj, 'collection'):
+            pk = obj.collection.namespace.pk
+            res.update({
+                'namespace': reverse('api:namespace_detail', kwargs={'pk': pk})
+            })
+        if getattr(obj, 'repository') is None:
+            return res
```
ironfroggy: if this block and the original block below now deal with two very different kinds of content, i think it warrants some comments detailing the different responsibilities. Its not clear just reading the code that lines 357-358 are collection specific, or that this implies the rest of the method is legacy repo specific.

galaxy/api/serializers/serializers.py
```diff
@@ -363,7 +374,7 @@ def get_related(self, obj):
         return res
 
     def get_summary_fields(self, obj):
-        if obj is None:
+        if obj is None or getattr(obj, 'repository') is None:
```
ironfroggy: Why are there no summary fields for collection importing?

galaxy/pulp/tasks.py
```diff
@@ -25,63 +25,57 @@
...
+            coll, coll_ver = _publish_collection(artifact, repository,
+                                                 namespace_pk, collection_info)
+        _process_import_success(coll, coll_ver, import_task)
+    except i_exc.ImporterError as e:
+        _process_import_fail(artifact, import_task, msg=e)
+    except exc.VersionConflict:
+        msg = 'Collection version already exists in galaxy'
+        _process_import_fail(artifact, import_task, msg)
+    except Exception as e:
+        _process_import_fail(artifact, import_task, msg=e.__class__.__name__)
```
ironfroggy: It looks like this will drop a lot more information, like the `msg` of the original exception. Can that be preserved?

galaxy/api/serializers/serializers.py
awcrosby: I will clean this up and make it more clear one block is for collections and the other repositories

galaxy/api/serializers/serializers.py
awcrosby: These edits are to get collection imports storing import task messages, `api/v1/imports` edits are temporary (right now to verify task logging) until `api/v2/collection` views are in place

galaxy/pulp/tasks.py
awcrosby: On repository import an uncaught error says only “Task # failed”, here we give a bit more info (error name) but wanted to continue to avoid providing the galaxy stacktrace to the enduser.

ironfroggy: If the information gets clipped here to keep it from the user, can we at least log the exception data for ourselves?

awcrosby: Ah yes, this wasn't getting into our pulp logs, thanks for the catch

galaxy/main/migrations/0130_import_task_collection_support.py
```diff
@@ -0,0 +1,37 @@
...
+
+    dependencies = [
+        ('main', '0129_collection_quality_score_updates'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='importtask',
+            name='artifact_id',
+            field=models.CharField(blank=True, max_length=100, null=True),
```
cutwater: Why artifact_id has `CharField` type?

galaxy/main/models/importing.py
```diff
@@ -18,10 +18,24 @@ class Meta:
     STATE_FAILED = constants.ImportTaskState.FAILED.value
     STATE_SUCCESS = constants.ImportTaskState.SUCCESS.value
 
+    # collection related
+    artifact_id = models.CharField(
```
cutwater: ditto

galaxy/main/migrations/0130_import_task_collection_support.py
awcrosby: I had referenced the celery_task_id being CharField, but can change to `IntegerField`

galaxy/main/models/importing.py
```diff
@@ -18,10 +19,27 @@ class Meta:
     STATE_FAILED = constants.ImportTaskState.FAILED.value
     STATE_SUCCESS = constants.ImportTaskState.SUCCESS.value
 
+    # collection related
+    artifact = models.ForeignKey(
```
cutwater: How artifact reference in this model is planned to be used?

awcrosby: @cutwater I wanted to associate the import task on creation to something related to collection. But haven't needed artifact reference and it is not persistent in database, I am thinking it is better to remove it

alikins:
looks good to me, testing with a 'mazer publish'. I end up with an entry in /api/v1/imports for the collection import.

PR #1639, opened 2019-03-13
Add collection views to `galaxy.api.v2` package

cutwater:


alikins:
The API version introspection bits don't show info about v2 yet.

The 'available_versions' on galaxy.api.views.views.ApiRootView
only shows 'v1' but should presumably show 'v2' as well. 

Not sure what 'current_version' should show in this scenario. It
previously was always 'v1', but maybe now should be 'v2'?

mazer checks if the value of 'current_version' is in it's SUPPORTED_VERSIONS
list. The SUPPORTED_VERSIONS currently is just 'v1', but mazer will get patched
for that shortly to be '['v1', 'v2'].


The response for 'GET /api'  (galaxy.api.views.views.ApiRootView) shows:

``` json
{
    "description": "GALAXY REST API",
    "current_version": "v1",
    "available_versions": {
        "v1": "/api/v1/"
    },
    "server_version": "3.0.3.dev587+gf4dce0b",
    "version_name": "Doin' it Right",
    "team_members": [
        "chouseknecht",
        "cutwater",
        "alikins",
        "newswangerd",
        "awcrosby",
        "tima",
        "gregdek"
    ]
}
```

alikins:
With some of the mazer tweaks mentioned above, 'mazer publish' that uses 'POST /api/v2/collections' seems to work.

alikins:
For /api/v2,  should /api/v2 also contain all of the /api/v1 endpoints?  So api/v2/ is a superset of
both apis (or at least, everything in v1 that isnt deprecated).

Can I assume /api/v2/collections will never appear in a 'stable' release?

alikins:
https://github.com/ansible/mazer/pull/199 has a mazer PR that updates collection publishing to using the /api/v2/collections endpoint added in this pr.

cutwater:
> The API version introspection bits don't show info about v2 yet.

I haven't touched our API root view for reason, we had couple of ideas what to do with root view, but I think nothing have been decied.

> For /api/v2, should /api/v2 also contain all of the /api/v1 endpoints? So api/v2/ is a superset of
both apis (or at least, everything in v1 that isnt deprecated).

No, `/api/v2` will never include `/api/v1`. APIv2 is planned to use different approach and drop all legacy bad practices we used in APIv1.

> Can I assume /api/v2/collections will never appear in a 'stable' release?

Please elaborate.

alikins:
@cutwater 

> Can I assume /api/v2/collections will never appear in a 'stable' release?

Doh, sorry, typo. I meant '/api/v1/collections'.  I'm assuming that the /api/v1/collections that existed in devel/ was devel only and only temporary. Released 'stable' versions will use '/api/v2/collections'

Issue #1661, opened 2019-03-20
Collections: Implement API v2 endpoints


cutwater:
Implement required API endpoints for collections according to the #1653 

### Action items:

- [x] Implement collection import task detail endpoint #1662
- [x] Collection detail API endpoint #1727
- [x] Collection version list API endpoint #1728
- [x] Artifact download API endpoint #1744

PR #1662, opened 2019-03-20
Implement collection import task detail endpoint

cutwater:
Issue: #1661

### DIFF
--- a/galaxy/main/models/collection.py
+++ b/galaxy/main/models/collection.py
@@ -89,9 +89,9 @@ class CollectionVersion(mixins.TimestampsMixin, pulp_models.Content):
     # References
     collection = models.ForeignKey(
         Collection, on_delete=models.PROTECT, related_name='versions')
-    _content = models.OneToOneField(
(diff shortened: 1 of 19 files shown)


galaxy/main/models/collection.py
```diff
@@ -89,9 +89,9 @@ class CollectionVersion(mixins.TimestampsMixin, pulp_models.Content):
     # References
     collection = models.ForeignKey(
         Collection, on_delete=models.PROTECT, related_name='versions')
-    _content = models.OneToOneField(
-        pulp_models.Content, on_delete=models.CASCADE, parent_link=True,
-        related_name='+', db_column='id',
+    _parent_ref = models.OneToOneField(
```
ironfroggy: this says `_parent_ref` but the migration says `parent_ref`, no underscore.

cutwater: It is `_parent_ref` with underscore in both cases. Am I missing something?

ironfroggy: huh... obviously i was _really_ sure i did not see an underscore there yesterday, but clearly i cannot read.

Issue #1643, opened 2019-03-14
Enable management of Collections on My Content


newswangerd:
- [x] Add API to load lists of collections
- [x] Add toggle to switch between repos and collections
- [x] Refactor page to allow handling of repos and collections
- [x] Set upload button to load modal for uploading new version of a collection
- [x] Refactor page as react component
- [x] ~~Update list expansion to work with collections~~ (dropping for now)
- [x] Add import status and links to import page (blocked: need import API)
- [x] Hook up deprecation option to API (blocked: need API for updating collections)
- [x] Add link to collection detail page (blocked: need to build collection detail page)

PR #1672, opened 2019-03-22
Enable collections on my-content.

newswangerd:
#1643 

Display Repositories
Screen Shot 2019-03-25 at 11 24 50 AM

Display Collections
Screen Shot 2019-03-25 at 11 24 58 AM

Upload new collection version
Screen Shot 2019-03-25 at 11 25 43 AM

Additional Changes
- Removed the dropdown view from repositories that showed warnings. This feature hasn't made it to production yet and having an expandable list inside of an expandable list is a bad user experience. Showing warnings will be the responsibility of the My Imports page.
- Namespace content view has been migrated to react and refactored to be easier to maintain.

Things that will be added once their corresponding data models/views are available:
- Redirects from the add content modal to the import page. Right now the modal just closes after a collection is imported and does nothing. Once the collection import page is available it should redirect to that.
- Links to the collection detail page. Right now they route to `/collection/{namespace}/{name}` which currently displays a 'content not found warning'
- Links to import view. Right now this just links to the import page for a repository, but this will be updated to link to the collections import once it's available.
- Deprecation option doesn't have an API endpoint yet so it doesn't do anything.


galaxy/api/internal/serializers/__init__.py
```diff
@@ -14,3 +14,5 @@
 #
 # You should have received a copy of the Apache License
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+from .collections import *              # noqa
```
cutwater: Please replace it with explicit list of imported names.

```
from .collections import (  # noqa: F401
    CollectionListSerializer,
)
```

galaxy/api/internal/serializers/collections.py
```diff
@@ -0,0 +1,58 @@
...
+            'namespace',
+            'name',
+            'deprecated',
+            'download_count',
+            'community_score',
+            'latest_version',
+        )
+
+    def get_latest_version(self, obj):
+        latest_version = models.CollectionVersion.objects.filter(
```
cutwater: Better to implement it as `CollectionVersion` model method.

galaxy/api/internal/urls.py
```diff
@@ -17,9 +17,15 @@
 
 # from django.urls import path
 
-# from galaxy.api.internal import views
+from galaxy.api.internal import views
+from django.conf.urls import url
```
cutwater: Django regexp urls are about to be deprecated. Please use `from django.urls import path`

galaxy/api/internal/views/__init__.py
```diff
@@ -14,3 +14,5 @@
 #
 # You should have received a copy of the Apache License
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+from .collections import *              # noqa
```
cutwater: ditto

galaxy/main/migrations/0132_auto_20190322_1528.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.1.5 on 2019-03-22 19:28
```
cutwater: Please don't leave auto-generated migration names.

PR #1673, opened 2019-03-22
Collection filename validation against manifest

awcrosby:


galaxy/api/v2/views/collection.py
```diff
@@ -76,6 +76,7 @@ def post(self, request, *args, **kwargs):
                 'repository_pk': repository.pk,
                 'namespace_pk': namespace.pk,
                 'task_id': import_task.id,
+                'filename': data['filename'],
```
awcrosby: @cutwater I didn't see a way to access filename from artifact object, so passed it forward here

galaxy/importer/collection.py
```diff
@@ -100,6 +104,16 @@ def _load_collection_readme(self):
                 'Readme listed in manifest not found: '
                 '{}'.format(self.collection_info.readme))
 
+    def _check_filename_matches_manifest(self):
+        metadata = self.collection_info
+        f = self.filename
+        if f.namespace != metadata.namespace or f.name != metadata.name:
+            raise exc.ManifestValidationError('Filename did not '
```
cutwater: IMO: Wrapping arguments is more readable:

```
raise exc.ManifestValidationError(
    'Filename did not match metadata')
```

galaxy/api/v2/views/collection.py
cutwater: It's ok

Issue #1624, opened 2019-03-05
Redirect URLs for /{namespace}/{repo}


newswangerd:
We've decided not to enforce unique `{namespace}.{content_name}` across legacy roles and collections. To support this on the front end we need to update our routing with the following:

- [x] Collections details will be routed via `/collection/{namespace}/{name}`
- [x] Legacy roles will be routed via `/repository/{namespace}/{name}`
- [x] `/{namespace}/{name}` will be redirected to `/repository/{namespace}/{name}` to preserve existing URL routes
- [x] `/collection` and `/repository` will redirect to `/search`
- [x] `/collection/{namespace}` and `/repository/{namespace}` will route to `/{namespace}`

PR #1680, opened 2019-03-25
Update URL schema to support collections.

newswangerd:
Fixes: #1624

- Collections details routed via `/collections/{namespace}/{name}`
- Legacy roles routed via `/repositories/{namespace}/{name}`
- `/{namespace}/{name}` redirects to `/repositories/{namespace}/{name}` to preserve existing URL routes
- `/collections` and `/repositories` redirect to `/search`
- `/collections/{namespace}` and `/repositories/{namespace}` route to `/{namespace}`
- Adds a placeholder component for collection details


alikins:
I'm kind of undecided on 'always plural' vs 'always singular'. The URLs end up feeling odd sometimes either way. ie `GET /collection` returning a list feels odd, but so does `GET /collections/37` returning a single item.

https://www.moesif.com/blog/api-guide/api-design-guidelines/#plural-vs-singular-resource-names has some references for different viewpoints.

PR #1686, opened 2019-03-26
Decouple score calculation from database

cutwater:

### DIFF
--- a/galaxy/importer/loaders/base.py
+++ b/galaxy/importer/loaders/base.py
@@ -21,87 +21,15 @@ import os
 
 from galaxy.common import logutils
 from galaxy.importer.utils import readme as readmeutils
-from galaxy.main import models
+from galaxy.importer.utils import lint as lintutils
 
 
 default_logger = logging.getLogger(__name__)
-
-BASE_SCORE = 50.0
-SEVERITY_TO_WEIGHT = {
-    0: 0.0,
-    1: 0.75,
-    2: 1.25,
-    3: 2.5,
-    4: 5.0,
-    5: 10.0,
-}
-CONTENT_SEVERITY_TYPE = 'content'
-CONTENT_SEVERITY = {
-    'ansible-lint_e101': 3,
-    'ansible-lint_e102': 4,
-    'ansible-lint_e103': 5,
-    'ansible-lint_e104': 5,
-    'ansible-lint_e105': 4,
-    'ansible-lint_e201': 0,
-    'ansible-lint_e202': 5,
-    'ansible-lint_e203': 2,
-    'ansible-lint_e204': 1,
-    'ansible-lint_e205': 3,
-    'ansible-lint_e206': 2,
-    'ansible-lint_e301': 4,
-    'ansible-lint_e302': 5,
-    'ansible-lint_e303': 4,
-    'ansible-lint_e304': 5,
-    'ansible-lint_e305': 4,
-    'ansible-lint_e306': 3,
-    'ansible-lint_e401': 3,
-    'ansible-lint_e402': 3,
-    'ansible-lint_e403': 1,
-    'ansible-lint_e404': 4,
-    'ansible-lint_e501': 5,
-    'ansible-lint_e502': 3,
-    'ansible-lint_e503': 3,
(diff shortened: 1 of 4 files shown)


galaxy/importer/loaders/role.py
```diff
@@ -172,73 +172,74 @@ def parse_videos(self):
...
-                          ', '.join(tested_versions)))
-
-        untested = set(supported_versions) - set(tested_versions)
-        if untested:
-            self.log.info('Not all supported versions tested: '
-                          '{}'.format(', '.join(untested)))
-            return False
-
-        return True
+    # NOTE(cutwater): This function currently not used.
```
cutwater: Existing scoring approach will not work for RoleMetaParser.
It seems unused therefore it's temporarily commented.

galaxy/importer/loaders/base.py
```diff
@@ -166,14 +167,36 @@ def lint(self):
...
 
-    def _on_lint_issue(self, linter_type, rule_id, message):
-        extra = {
-            'is_linter_rule_violation': True,
-            'linter_type': linter_type,
-            'linter_rule_id': rule_id,
-            'rule_desc': message
+    def score(self):
+        content_w = self._score_stats.get(CONTENT_SEVERITY_TYPE, 0.0)
+        content_score = max(0.0, (BASE_SCORE - sum(content_w)) / 10)
```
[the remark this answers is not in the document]
cutwater: Fixed already.

galaxy/importer/loaders/role.py
cutwater: Ok, will drop it.

galaxy/importer/loaders/base.py
```diff
@@ -166,14 +167,36 @@ def lint(self):
 
         return all_linters_ok
 
-    def _on_lint_issue(self, linter_type, rule_id, message):
-        extra = {
-            'is_linter_rule_violation': True,
-            'linter_type': linter_type,
-            'linter_rule_id': rule_id,
-            'rule_desc': message
+    def score(self):
```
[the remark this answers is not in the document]
cutwater: Yep, actually it should go to RoleLoader.

galaxy/importer/loaders/base.py
```diff
@@ -166,14 +95,13 @@ def lint(self):
...
-            'linter_type': linter_type,
-            'linter_rule_id': rule_id,
-            'rule_desc': message
-        }
-        self.log.warning(message, extra=extra)
+    def score(self):
+        return None
+
+    def _on_lint_issue(self, linter_type, rule_id, rule_desc, message=None):
+        lint_record = lintutils.LintRecord(linter_type, rule_id, rule_desc)
```
alikins: Looks good to me.

(The comments below are about a related idea not directly related to this pr, but making a note while it is in my head...)

I might be tempted in the future to try to batch up the lint result log message logging / ImportTaskMessage creationion a bit, but there would be a premature optimization until its needed. Afaict, since the ImportTaskMessage table doesn't have hard deps on other tables there shouldn't be any issue with creating a lot of them and/or at high rate (ie, there doesn't appear to be any reason at the moment why batching would be needed)

Could the LintRecord creation and the `self.log.warning(message, extra={'lint_record': lint_record})` happen in separate steps?  One method to create/populate the LintRecord and return it, and then send it to the logger elsewhere?

If those are separate, I'm wondering if in the `lint()` method, if all of the LintRecords could be accumulated and sent in a batch as each step finishes (or as smaller batches within each step). ie, wait for N number of LintRecords() to be created, then batch emit them to the logger. _I don't think the style of batching I mention is a necessary optimization currently and may well never be_ (but, again, just making notes while viewing the related code)

cutwater: > Could the LintRecord creation and the self.log.warning(message, extra={'lint_record': lint_record}) happen in separate steps? One method to create/populate the LintRecord and return it, and then send it to the logger elsewhere?

Yes, it's a good idea. I was thinking on the same approach. The main goal was to unblock collections implementations, so it may not be perfect. Let's update it in a separate patch.

PR #1689, opened 2019-03-27
Add code review comments

cutwater:


galaxy/importer/linters/__init__.py
```diff
@@ -68,16 +68,26 @@ def parse_id_and_desc(self, message):
             rule_desc = ' '.join(msg_parts[2:])
 
             error_id = msg_parts[1]
+            # REVIEW(cutwater): Flake8 is configured to return `E,F,W` rules.
+            #   Why is parsing rules prefixes list differs
+            #   from FLAKE8_SELECT_ERRORS.
```
awcrosby: In deciding which flake8 issues to score at which severity, we did not want to map and maintain by individual rule, so we used E and W since they implied categorical severity, while the F pyflakes did not. Due in part to this complexity, we eventually decided to postpone scoring via flake8 (importer/loaders/base.py severity mappings do not have any flake8 matching), and keep it for ansible-lint (mapped to individual rules) and yamllint (mapped to categorical errors and warnings)

galaxy/importer/linters/__init__.py
```diff
@@ -68,16 +68,26 @@ def parse_id_and_desc(self, message):
...
             if error_id[0] not in ['E', 'W']:
                 error_id = None
 
         except IndexError:
+            # REVIEW(cutwater): If this except clause is executed,
+            #  rule_desc may not be defined, but it's used below.
             error_id = None
```
awcrosby: If this except clause is executed, error_id should be None and the next code block should return before rule_desc is referenced

galaxy/importer/linters/__init__.py
```diff
@@ -68,16 +68,26 @@ def parse_id_and_desc(self, message):
...
 
         if not error_id:
-            logger.error('No error_id found in {} message'.format(self.cmd))
+            # REVIEW(cutwater): Couple of questions here:
+            #   1. Why is that error?
+            #   2. Why displaying command that was executed
+            #      but not the message that is parsed?
+            logger.error('No error_id found in message: {}'.format(self.cmd))
```
awcrosby: This is mainly for debugging in case linter outputs a line in a different format, and yes it should be `message` logged.

galaxy/importer/linters/__init__.py
cutwater: Indeed, however the code flow is fuzzy and it's recognized by static code analysis tools as a potential issue.
You could've invert the logic and return error_id and message if it's found and handle error case down below.

```
        try:
            error_id, rule_desc = message.split(' ', 1)
        except ValueError:
            pass
        else:
            if error_id in ['E', 'W']:
                return error_id, rule_desc

        # exception is probably better here
        logger.warning('Cannot parse linter message: {}'.format(message))
        return None, None
```

awcrosby:
Comments addressed in #1751

Issue #1670, opened 2019-03-21
403 during access to owned roles


Voronenko:
## Bug Report 
Galaxy server returns 403 for some roles
"You do not have permission to perform this action"

##### SUMMARY
Expected scenario: 
1) role owner should be able to open details about his role
2) role description should be available via api call  https://galaxy.ansible.com/api/v1/content/

Actual result: some roles are not accessible, API returns 403 trying to access api/v1/content/15115/

##### STEPS TO REPRODUCE

I observe at least on 3 (older) roles

https://galaxy.ansible.com/softasap/sa-rsyslog-server
https://galaxy.ansible.com/api/v1/content/15115/?format=json

https://galaxy.ansible.com/softasap/sa-mysql-percona
https://galaxy.ansible.com/api/v1/content/15977/

https://galaxy.ansible.com/softasap/sa-mysql56
https://galaxy.ansible.com/api/v1/content/15977/

haven't checked all the list

##### EXPECTED RESULTS

Expected scenario: 
1) role owner should be able to open details about his role
2) role description should be available via api call  https://galaxy.ansible.com/api/v1/content/

##### ACTUAL RESULTS

some roles are not accessible, API returns 403 trying to access api/v1/content/<roleId>/

Voronenko:
Perhaps faulty component is ansible-lint , but that should not cause 403...

Voronenko:
@chouseknecht Can you take a look on POC above and perhaps suggest root cause?
Summarizing, in case of issues with current Metadata - api should not return 403, but provide more meaningful message

chouseknecht:
That's odd. https://galaxy.ansible.com/api/v1/content/?id=15115 seems to work OK.

chouseknecht:
This one works fine: https://galaxy.ansible.com/api/v1/content/22437/

Voronenko:
yes, but `?format=json`  used by galaxy web UI and prevents from rendering.
I am pretty sure it has to do with ansible linting itself.

awcrosby:
@Voronenko it looks like we are expecting the `meta/main.yml` `galaxy_tags` to be string and not numeric. We should be checking for this on our side, we will get it fixed, but in the meantime you can put quotes around the numeric tags in those 3 roles they should import successfully

awcrosby:
Going ahead and closing this, if you still have issues after you change all your tags to strings, please let us know

Voronenko:
[...]

PR #1690, opened 2019-03-27
Ignore role tags when not strings

awcrosby:
Signed-off-by: Andrew Crosby <acrosby@redhat.com>

### DIFF
--- a/galaxy/importer/loaders/role.py
+++ b/galaxy/importer/loaders/role.py
@@ -81,6 +81,8 @@ class RoleMetaParser(object):
         return dependencies
 
     def _validate_tag(self, tag):
+        if not isinstance(tag, str):
+            return False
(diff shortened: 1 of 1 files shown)


galaxy/importer/loaders/role.py
```diff
@@ -81,6 +81,8 @@ def _get_dependencies(self, metadata):
         return dependencies
 
     def _validate_tag(self, tag):
+        if not isinstance(tag, str):
+            return False
```
alikins: Looks goods to me.

I was curious how/if the collection import models and validators handle it (galaxy.importer.models.CollectionInfo) and it looks like the validators there already handle it.

Code as is in this PR looks good to me.

awcrosby:
Prevents importer error when tags are numeric as noted in #1670

PR #1692, opened 2019-03-27
Support license as list and license_file in collection manifest

awcrosby:
Incorporates applicable license changes made in https://github.com/ansible/mazer/pull/207

### DIFF
--- a/galaxy/importer/models.py
+++ b/galaxy/importer/models.py
@@ -93,7 +93,7 @@ class BaseCollectionInfo(object):
     namespace = attr.ib(default=None)
     name = attr.ib(default=None)
     version = attr.ib(default=None)
-    license = attr.ib(default=None)
+    license = attr.ib(factory=list)
     description = attr.ib(default=None)
 
     # TODO: see how to handle license_list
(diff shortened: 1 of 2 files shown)


galaxy/importer/tests/test_collection_info.py
```diff
@@ -86,18 +86,18 @@ def galaxy_col_info(base_col_info):
...
 
 valid_licenses = [
-    'MIT',
-    'Apache-2.0',
-    'BSD-3-Clause',
-    'CC-BY-4.0'
+    ['MIT'],
+    ['Apache-2.0'],
+    ['BSD-3-Clause'],
+    ['CC-BY-4.0', 'MIT']
```
alikins: +1 For included an intentional dupe in the test data list.  Good way to catch potential assumptions about data ordering and uniqueness.

For this particular case, in theory the valid_licenses is supposed to be an arbitrarily ordered (but usually sorted) list derived from the keys of a dict. But it is easy for the app code to have bugs because of assumptions about the data, so good thing to test for.

galaxy/importer/models.py
```diff
@@ -88,26 +88,31 @@ def __init__(self, branch, commit, format, contents,
...
+    '''Represents collection_info metadata in collection manifest.
+
+    Includes data validation expected when collection is built.
+    '''
 
     namespace = attr.ib(default=None)
     name = attr.ib(default=None)
     version = attr.ib(default=None)
-    license = attr.ib(default=None)
+    license = attr.ib(factory=list)
```
alikins: (note for potential later discussion) I wouldn't mind changing 'license' attr to 'licenses' but that is something we can discuss later and in the collection spec. Reasons being a) it is a list b) the python builtin 'license' name conflict

alikins:
Looks good to me, and testing with mazer publish seems to work. :+1:

Issue #1660, opened 2019-03-20
Store tasks parameters and data


cutwater:
Pulp's tasking system doesn't allow storing task parameters, result and additional metadata in the task object.

Implement capability to store that data in associated with tasks database table.

PR #1699, opened 2019-03-28
Persistent task parameters and collection import data

cutwater:
Closes: #1660


galaxy/common/tasking.py
```diff
@@ -0,0 +1,62 @@
...
+
+    :return: Created Task object.
+    """
+    resources = resources or []
+    job = enqueue_with_reservation(
+        func=func, resources=resources or [],
+        kwargs=params, options=options)
+    pulp_task = pulp_models.Task.objects.get(job_id=job.id)
+    return task_cls.objects.create(
+        params=params, pulp_task=pulp_task, **task_args)
```
ironfroggy: Should these lines be atomic? If the `task_cls` creation fails, will it leave an orphaned Pulp task that isn't being tracked or doesn't have associated data we're keeping in our model?

galaxy/common/tasking.py
```diff
@@ -0,0 +1,62 @@
...
+def save_result(func):
+    """Saves value returned by task to the database."""
+    @functools.wraps
+    def wrapper(*args, **kwargs):
+        result = func(*args, **kwargs)
+
+        task = models.Task.current()
+        task.result = result
+        task.save()
+    return wrapper
```
ironfroggy: This does not seem to be used anywhere.

galaxy/worker/logutils.py
```diff
@@ -29,8 +28,8 @@ def process(self, msg, kwargs):
 
 
 class ImportTaskAdapter(BaseLoggerAdapter):
-    def __init__(self, logger, task_id):
-        super(ImportTaskAdapter, self).__init__(logger, {'task_id': task_id})
+    def __init__(self, logger, task):
+        super(ImportTaskAdapter, self).__init__(logger, {'task': task})
```
ironfroggy: In Python 3 you can omit the arguments to `super` and just do
```         super().__init__(logger, {'task': task}) ```

see https://www.python.org/dev/peps/pep-3135/

File: galaxy/worker/logutils.py [hunk 4]
```diff
@@ -62,9 +61,22 @@ def emit(self, record: logging.LogRecord) -> None:
...
+            message_type=const.ImportTaskMessageType.from_logging_level(
                 record.levelno).value,
             message_text=record.msg,
             **create_kwargs,
         )
+
+
+class CollectionImportHandler(logging.Handler):
+    def emit(self, record: logging.LogRecord) -> None:
+        from galaxy.main import models
```
ironfroggy: Is this local import absolutely necessary?

galaxy/common/tasking.py
cutwater: These are atomic. Check decorator on line 27.

galaxy/common/tasking.py
cutwater: Yes, I initially planned to use it to save task results, which is still an option. 
But for `import_collection` i'm creating reference to a created version via `ForeignKey`, as it'll be used to query task for version in reverse direction.

galaxy/worker/logutils.py
cutwater: Sure, will update. Dropping python 2 compatible code was not a part of this PR. I may change it here and in other places as a separate PR or update this one.

galaxy/worker/logutils.py
cutwater: Currently it's necessary, because handlers are initialized as a part of logging config, which happens in settings and at that time django is not initialized and models imports at the module level fails.
I have an idea how to refactor this and move logging initialization for import tasks out of settings, but probably won't happen as this PR. I'll add TODO comment to refactor it in future.

galaxy/common/tasking.py
ironfroggy: overlooked that, my bad

galaxy/common/tasking.py
ironfroggy: Right, so if you ended up doing something else, shouldn't this unused, unreachable function be removed from the diff?

galaxy/worker/logutils.py
ironfroggy: I'm a big fan of "as you go" updates and refactors, which is why I only said anything at all because this line was already being touched. Its not a big deal, just pointing it out in-case the change wasn't on your radar. Not a blocker.

galaxy/worker/logutils.py
ironfroggy: That's a fine reason. Certainly have hit that exact same limitation myself, numerous times. 👌

galaxy/api/v2/views/collection.py
```diff
@@ -65,22 +66,19 @@ def post(self, request, *args, **kwargs):
...
-                'filename': data['filename'],
+        task = tasking.create_task(
+            tasks.import_collection,
+            task_cls=models.CollectionImport,
+            params={
+                'artifact_id': artifact.pk,
+                'repository_id': repository.pk,
+            },
+            task_args={
+                'namespace': namespace,
```
newswangerd: Are you planning on adding API endpoints to query imports by namespace as part of this PR, or is that coming later?

cutwater: Yes, working on this in terms of this PR.

galaxy/common/tasking.py
cutwater: It's part of the library. It is reachable and will be useful for storing task results that don't have standalone database entities.

galaxy/api/v2/views/collection_import.py
```diff
@@ -31,7 +31,7 @@ class CollectionImportView(views.APIView):
     permission_classes = (IsAuthenticated, )
 
     def get(self, request, *args, **kwargs):
-        task = get_object_or_404(pulp_models.Task, pk=self.kwargs['pk'])
+        task = get_object_or_404(models.Task, pk=self.kwargs['pk'])
```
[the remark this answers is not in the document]
cutwater: Fixed

galaxy/api/v2/views/collection.py
cutwater: Done

galaxy/main/models/importing.py
```diff
@@ -18,18 +18,11 @@ class Meta:
     STATE_FAILED = constants.ImportTaskState.FAILED.value
     STATE_SUCCESS = constants.ImportTaskState.SUCCESS.value
 
-    collection = models.ForeignKey(
```
newswangerd: Importing repositories breaks and I suspect it's because this field was removed. I think the collection field needs to be removed from the importtask serializer.

```
2019-04-02 14:56:18,651 ERROR django.request: Internal Server Error: /api/v1/imports/
Traceback (most recent call last):
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/django/core/handlers/base.py", line 126, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/django/core/handlers/base.py", line 124, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/rest_framework/views.py", line 494, in dispatch
    response = self.handle_exception(exc)
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/rest_framework/views.py", line 454, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/rest_framework/views.py", line 491, in dispatch
    response = handler(request, *args, **kwargs)
  File "/galaxy/galaxy/api/views/views.py", line 440, in post
    response = {'results': [serializer.data]}
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/rest_framework/serializers.py", line 537, in data
    ret = super(Serializer, self).data
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/rest_framework/serializers.py", line 262, in data
    self._data = self.to_representation(self.instance)
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/rest_framework/serializers.py", line 487, in to_representation
    fields = self._readable_fields
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/django/utils/functional.py", line 37, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/rest_framework/serializers.py", line 377, in _readable_fields
    field for field in self.fields.values()
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/rest_framework/serializers.py", line 363, in fields
    for key, value in self.get_fields().items():
  File "/galaxy/galaxy/api/serializers/serializers.py", line 94, in get_fields
    ret = super(BaseSerializer, self).get_fields()
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/rest_framework/serializers.py", line 1022, in get_fields
    source, info, model, depth
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/rest_framework/serializers.py", line 1167, in build_field
    return self.build_unknown_field(field_name, model_class)
  File "/usr/share/galaxy/venv/lib64/python3.6/site-packages/rest_framework/serializers.py", line 1279, in build_unknown_field
    (field_name, model_class.__name__)
django.core.exceptions.ImproperlyConfigured: Field name `collection` is not valid for model `ImportTask`.
2019-04-02 14:56:18,656 ERROR django.server: "POST /api/v1/imports/ HTTP/1.1" 500 24867
```

galaxy/api/internal/views/imports.py
```diff
@@ -0,0 +1,97 @@
...
+        ),
+    }
+
+    def list(self, request, *args, **kwargs):
+        items = self.paginate_queryset(self.get_queryset())
+        self.load_objects(items)
+        result = self.serialize_objects(items)
+        return self.get_paginated_response(result)
+
+    def get_queryset(self):
```
newswangerd: Can you add parameters to filter by type (collection | repository) and name?

cutwater: Sure

galaxy/main/models/task.py
```diff
@@ -38,6 +38,30 @@ class Task(models.Model):
         related_name='galaxy_task'
     )
 
+    @property
+    def job_id(self):
+        return self.pulp_task.job_id
+
+    @property
+    def state(self):
+        return self.pulp_task.state
```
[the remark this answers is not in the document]
cutwater: Let's map old statuses to pulp ones.

galaxy/worker/tasks/collection.py
```diff
@@ -0,0 +1,143 @@
...
+        namespace=task.namespace, name=metadata.name)
+
+    try:
+        version = collection.versions.create(
+            version=metadata.version,
+            metadata=metadata.get_json(),
+            quality_score=collection_info.quality_score,
+        )
+    except IntegrityError:
+        raise exc.VersionConflict('Collection version already exists.')
```
alikins: Since the code has the metadata here, the error message could include more details like the metadata.version, namespace, name etc.

Don't know if it would be worthwhile right now to add some of that info as attributes of VersionConflict. But maybe in the future.

galaxy/worker/tasks/collection.py
```diff
@@ -0,0 +1,143 @@
...
+
+    relative_path = '{0}-{1}-{2}.tar.gz'.format(
```
alikins: Could use something like

``` python
relative_path = '{metadata.namespace}-{metadata.name}-{metadata.version}.tar.gz'.format(metadata=metadata)
```

Not sure if it matters much for this case, so either is fine.

galaxy/worker/tasks/collection.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def _process_collection(artifact, filename, task_logger):
+    with tempfile.TemporaryDirectory() as extract_dir:
+        with artifact.file.open() as pkg_file, \
+                tarfile.open(fileobj=pkg_file) as pkg_tar:
+            pkg_tar.extractall(extract_dir)
+
+        try:
+            collection_info = importer.import_collection(
```
alikins: ~~The combo of `from galaxy.importer import collection as importer` and `collection_info = import.import_collection()` is a little confusing to me. Especially In combo with the `from galaxy.importer import collection as importer` returns a `CollectionLoader`. 
And the var here is `collection_info` and `collection_info.collection_info.` is used in _log_collection_info.~~

I think the code does the right thing, just a little hard to follow the renames.

alikins: Nevermind, I missed the `load()` in `import.import_collection().load()`

galaxy/worker/tasks/collection.py
```diff
@@ -0,0 +1,143 @@
...
+        defaults={'publication': publication},
+    )
+
+    task.imported_version = version
+    task.save()
+
+
+def _log_collection_info(collection_info):
+    log.debug('Collection loaded - metadata={}, quality_score={}'.format(
+        collection_info.collection_info.__dict__,
```
alikins: I would be little hesitant to use `collection_info.__dict__` in the string formatting here, just because __dict__ could potentially end up with values that have heavy or finicky str/repr's. 

I think with this style of string formatting the formatting is done before the log.debug() call, the str/repr's for collection_info.__dict__ will always run, even for log levels below DEBUG.

As opposed to something like:

``` python
log.debug('Collection loaded - metadata=%s, quality_score=%s', collection_info.__dict__, collection_info.quality_score)
```

But if the goal is to render the string message now (as opposed to some later point in the logging stack), I think the code as is works.

awcrosby: These `log.debug` messages were added just for debugging locally and analysis on galaxy-dev, so using %-style strings would be fine

galaxy/worker/tasks/collection.py
awcrosby: Maybe a rename `importer_obj = importer.import_collection()` and therefore `metadata = importer_obj.collection_info`

galaxy/worker/logutils.py
```diff
@@ -62,9 +61,22 @@ def emit(self, record: logging.LogRecord) -> None:
...
+        task: models.CollectionImport = record.task
+
+        lint_record: lintutils.LintRecord = getattr(
+            record, 'lint_record', None)
+        if lint_record is not None:
+            task.add_lint_record(lint_record)
+        task.add_log_record(record)
```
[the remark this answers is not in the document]
cutwater: It's not str'fied because it's passed and processed as an object in `add_log_record` function.

galaxy/worker/tasks/collection.py
cutwater: Agree

galaxy/worker/tasks/collection.py
cutwater: Ok

Issue #1656, opened 2019-03-18
Check database dependent items on collection import


awcrosby:
- [ ] Check dependencies - for each dependent collection validate that namespace exists, name exists, version specifier is valid per `semantic_version`, and there exists a version of the collection that matches version specifier
- [ ] Check tags - validate tags are of the correct format, then write / replace in the Collection table
- [ ] Platforms / Cloud Platforms - TBD

PR #1700, opened 2019-03-29
Process tags at collection-level on import

awcrosby:
Works #1656

### DIFF
--- a/galaxy/main/models/collection.py
+++ b/galaxy/main/models/collection.py
@@ -24,6 +24,7 @@ from django.contrib.postgres import search as psql_search
 from django.core.validators import MinValueValidator, MaxValueValidator
 from django.db import models
 from pulpcore.app import models as pulp_models
+import semantic_version
 
 from galaxy.importer.utils import lint as lintutils
 from . import mixins
@@ -69,9 +70,20 @@ class Collection(mixins.TimestampsMixin, models.Model):
 
     @property
     def latest_version(self):
-        return CollectionVersion.objects.filter(
-            collection=self
-        ).latest('pk')
+        versions = self.versions.filter(hidden=False)
+        if not versions:
+            return None
+        return versions.latest('pk')
+
+    @property
+    def highest_version(self):
+        versions = self.versions.filter(hidden=False)
+        if not versions:
+            return None
+
+        d = {semantic_version.Version(v.version): v for v in versions}
+        highest = semantic_version.Spec('*').select(d.keys())
+        return d[highest]
 
 
 class CollectionVersion(mixins.TimestampsMixin, pulp_models.Content):
--- a/galaxy/importer/models.py
+++ b/galaxy/importer/models.py
@@ -286,12 +286,21 @@ class GalaxyCollectionInfo(BaseCollectionInfo):
             self.value_error('Dependency found in galaxy but no matching '
                              'version found: %s %s' % (dep_col, ver_spec))
 
+    def _check_tags_count(self):
+        '''Checks tag count in metadata against max tags count constant'''
+        tags = getattr(self, 'tags')
+        if tags is not None and len(tags) > constants.MAX_TAGS_COUNT:
+            self.value_error(
+                'Expecting no more than %s tags in metadata' %
+                constants.MAX_TAGS_COUNT)
+
     def __attrs_post_init__(self):
         '''Additional galaxy checks called post init'''
         super().__attrs_post_init__()
         self._check_required('readme')
         self._check_required('authors')
         self._check_dependencies()
+        self._check_tags_count()
         for field in [
                         'description',
                         'repository',
--- a/galaxy/importer/tests/test_collection_info.py
+++ b/galaxy/importer/tests/test_collection_info.py
@@ -149,6 +149,17 @@ def test_invalid_tags(base_col_info):
         assert tag in str(exc)
 
 
+def test_max_tags(galaxy_col_info):
+    galaxy_col_info['tags'] = [str(i) for i in range(90, 110)]
+    res = GalaxyCollectionInfo(**galaxy_col_info)
+    assert '101' in res.tags
+
+    galaxy_col_info['tags'] = [str(i) for i in range(90, 111)]
+    with pytest.raises(ValueError) as exc:
+        GalaxyCollectionInfo(**galaxy_col_info)
+    assert 'tags in metadata' in str(exc)
+
+
 def test_valid_semver(base_col_info):
     for ver in valid_semver:
         base_col_info['version'] = ver
--- a/galaxy/main/tests/test_collection_model.py
+++ b/galaxy/main/tests/test_collection_model.py
@@ -1,0 +1,84 @@
+# (c) 2012-2019, Ansible
+#
+# This file is part of Ansible Galaxy
+#
+# Ansible Galaxy is free software: you can redistribute it and/or modify
+# it under the terms of the Apache License as published by
+# the Apache Software Foundation, either version 2 of the License, or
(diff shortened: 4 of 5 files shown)


galaxy/main/models/collection.py
```diff
@@ -68,6 +69,13 @@ def latest_version(self):
             collection=self
         ).latest('pk')
 
+    @property
+    def highest_version(self):
+        versions = CollectionVersion.objects.filter(collection=self)
```
cutwater: `self.versions`

galaxy/main/models/collection.py
```diff
@@ -68,6 +69,13 @@ def latest_version(self):
...
+        pairs = [(v, semver.Version(v.version)) for v in versions]
+        pairs.sort(key=lambda x: x[1], reverse=True)
+        return pairs[0][0]
```
cutwater: In theory collection can have zero versions.

galaxy/main/models/collection.py
cutwater: hidden versions are not taken into account.

galaxy/main/models/collection.py
```diff
@@ -68,6 +69,13 @@ def latest_version(self):
             collection=self
         ).latest('pk')
 
+    @property
+    def highest_version(self):
+        versions = CollectionVersion.objects.filter(collection=self)
+        pairs = [(v, semver.Version(v.version)) for v in versions]
```
cutwater: `return sorted(versions, key=lambda x: semver.Version(x.version))[0]`

galaxy/main/models/collection.py
```diff
@@ -21,6 +21,7 @@
 from django.core.validators import MinValueValidator, MaxValueValidator
 from django.db import models
 from pulpcore.app import models as pulp_models
+import semantic_version as semver
```
alikins: Whats the reason for importing `semantic_version` as `semver`?

Since we previously used `semver` but stopped, seeing code that uses `semver.Version` initially reads as a bug.
If there is no reason not to, my preference would be to just `import semantic_version` and use `semantic_version.Version()` below

galaxy/pulp/tasks.py
```diff
@@ -164,3 +166,23 @@ def _log_collection_loaded(coll_info):
...
+
+
+def _write_collection_tags(collection, version, i_collection):
+    '''Update tags at collection-level, only if highest version'''
+
+    log.debug('highest_version={}'.format(collection.highest_version.version))
+    if collection.highest_version != version:
+        return
+
+    for tag in i_collection.collection_info.tags:
```
cutwater: Better use bulk operations here.

galaxy/main/models/collection.py
```diff
@@ -68,6 +69,13 @@ def latest_version(self):
...
 
+    @property
+    def highest_version(self):
+        versions = CollectionVersion.objects.filter(collection=self)
+        pairs = [(v, semver.Version(v.version)) for v in versions]
+        pairs.sort(key=lambda x: x[1], reverse=True)
+        return pairs[0][0]
+
 
 class CollectionVersion(mixins.TimestampsMixin, pulp_models.Content):
```
alikins: Could/should CollectionVersion be using a `semantic_version.django_fields.VersionField` for it's version field?
https://python-semanticversion.readthedocs.io/en/latest/django.html

galaxy/main/models/collection.py
alikins: Could potentially provide a `semantic_version.Spec()` here and use it's `select()` method: https://python-semanticversion.readthedocs.io/en/latest/reference.html#semantic_version.Spec.select

For now, the `Spec()` used could just be '*', but using the Spec would offer a way to support other version restrictions.

galaxy/main/models/collection.py
awcrosby: We have it like this in a few other places, maybe left overs from updating only the imported package, I will change it here.

galaxy/main/models/collection.py
awcrosby: @cutwater I also updated `latest_version` to account for hidden versions and zero versions

galaxy/main/models/collection.py
awcrosby: `Spec()` was added as the selector

galaxy/pulp/tasks.py
awcrosby: @cutwater I put a code comment to add bulk operations, should occur for role tags also as the same code is used in both places

galaxy/main/models/collection.py
cutwater: `semver` is not only a module of the same name package, but a commonly used abbreviation to "Semantic version". We dropped `semver` package entirely and it's not present anywhere in our dependencies, so I think it's still acceptable to use `semver` just as an abbreviation to the "Semantic version" term.

galaxy/main/models/collection.py
cutwater: Please don't. `semantic_version` package is not compliant with the spec and hopefully gets replaced with better alternative soon.

galaxy/pulp/tasks.py
cutwater: Can you please make it part of this PR.

galaxy/main/models/collection.py
```diff
@@ -64,9 +65,20 @@ class Meta:
...
+        return versions.latest('pk')
+
+    @property
+    def highest_version(self):
+        versions = self.versions.filter(hidden=False)
+        if not versions:
+            return None
+
+        d = {semantic_version.Version(v.version): v for v in versions}
+        highest = semantic_version.Spec('*').select(d.keys())
```
cutwater: @alikins any particular reason you wanted to use `semantic_version.Spec('*')` instead  of simple comparison, which does literally the same?

galaxy/main/models/collection.py
alikins: Import 'semantic_version as semver' makes it unnecessarily more difficult to find actual use of the `semantic_version` module  (ie, 'git grep semantic_version' only finds the import).

galaxy/main/models/collection.py
alikins: Which spec? And in what ways? I don't remember finding any ways `semantic_version` is not compliant with the semver spec/

galaxy/main/models/collection.py
alikins: @cutwater Using a semantic_version.Spec() provides a way to filter or blacklist if we need to at some point. Also allows the use of Spec.select() which does the ordering itself.

We don't necessarily need the filter/blacklist support now, so it may be YAGNI.

galaxy/main/models/collection.py
cutwater: Collections spec.

galaxy/main/models/collection.py
cutwater: It's ok to use aliases in python. We use lots of aliases for example for `rest_framework`. You wouldn't find `rest_framework` usages by just grep-ing the source code.

galaxy/main/models/collection.py
cutwater: Eventually this property should be replaced with FK, to calculate it once during import, shouldn't it?

awcrosby: A `ForeignKey` reference might be the better way to do it, however if that particular `CollectionVersion` was flagged as hidden, we may need to update the `ForeignKey`

galaxy/pulp/tasks.py
awcrosby: The bulk operations have been added to this PR

Issue #1658, opened 2019-03-18
Process collection contents metadata on import


awcrosby:
Information about contents inside a collection need to be stored in galaxy so they can be retrieved when searching. ~~Galaxy will use the content metadata defined inside the collection `galaxy.yml` as the source of truth to display to the end user.~~ Galaxy will introspect the collection to determine content metadata.

- Use CollectionVersion.contents JSONField for storage/retrieval of contents, as well as the historical storage of content metadata for each version
- The CollectionVersion.contents JSONField will be attempted to satisfy search requirements, if it cannot then a table will be used for contents. If using a table:
  - Store new content rows for all content items and tie to the Collection
  - Flag all older content rows tied to the Collection. Later can decide whether to delete flagged rows on import, or later in batch

awcrosby:
Closed with #1707

PR #1707, opened 2019-04-02
Store serialized collection content on import

awcrosby:
Works #1658

### DIFF
--- a/galaxy/importer/collection.py
+++ b/galaxy/importer/collection.py
@@ -17,6 +17,7 @@
 
 import os
 import logging
+import json
 
 import semantic_version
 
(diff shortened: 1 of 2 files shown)


galaxy/importer/collection.py
```diff
@@ -144,12 +147,53 @@ def _load_contents(self, content_list):
...
+            def default(self, obj):
+                try:
+                    return json.JSONEncoder.default(self, obj)
+                except TypeError:
+                    pass
+
+                try:
+                    return obj.name.lower()
+                except AttributeError:
+                    return str(obj).lower()
```
ironfroggy: This class isn't going to stay nested in here, is it?

ironfroggy: (WIP, I know, just asking!)

awcrosby: yeah i should move alongside the parent class or into utils

Issue #1620, opened 2019-03-05
Refactor existing models to work with collections


newswangerd:
Surveys and user preferences both have foreign key dependencies on repositories. We need to add similar dependencies for collections

Preferences
- [x] Add followed collections to `UserPreferences` model
  - [x] Update API with collections followed
  - [x] Update User preferences page
  - [x] Update content detail view follow functionality

Surveys
- [x] Add foreign key for collections to community survey
  - [x] Add `CollectionSurvey` endpoint
  - [x] Refactor community survey on content detail to handle surveys for collections
  - [x] Refactor user notifications

chouseknecht:
User notifications is being worked in #1803

PR #1708, opened 2019-04-02
Add support for collections in surveys and preferences

newswangerd:
#1620 

- Adds `CollectionsFollowed` to user preferences
- Adds `CollectionSurvey` model
- Renames `CommunitySurvey` to `RepositorySurvey`
- Sets survey urls to `api/v1/community_surveys/collection` and `api/v1/community_surveys/repository` for collections and repositories respectively.


galaxy/api/serializers/survey.py
```diff
@@ -42,22 +39,54 @@ class Meta:
...
+        model = models.RepositorySurvey
+        fields = SHARED_FIELDS + ('repository', )
+
+    def validate(self, data):
+        repo = data.get('repository')
+        user = data.get('user')
+
+        validate_not_owner(
+            user,
+            repo.provider_namespace.namespace.owners.all()
```
cutwater: Can you please use DB query for validating namespace owners.

PR #1714, opened 2019-04-04
[WIP] Adjust configuration for local test runs, and add first tests for internal UI API

ironfroggy:
Here's a starting point for unit tests against the internal UI API being added for collections.

It also includes some changes to make running tests easier, because the lower the friction on running tests the higher value the tests are going to have for the team. Here's a summary of the changes:

- Adding a make target `make test` to run tests locally, and moved some test configuration to allow running tests locally.
- Added make targets `make test/changed` and `make dev/test/changed` which only runs test modules that have uncommitted changes and skips coverage reporting (since they aren't valid when run on a subset of tests). This supports faster test runs and more efficient test/change/retest loops for the developer.
- Reduces most test run-times from minutes to 2 seconds.
- Coverage parameters are moved from `setup.cfg` to `Makefile` so they can be adjusted for the full versus partial targets

### DIFF
--- a/galaxy/api/v2/tests/test_collection_view.py
+++ b/galaxy/api/v2/tests/test_collection_view.py
@@ -1,0 +1,35 @@
+import pytest
+from django.test import TestCase
+from rest_framework.test import APIClient
+
+from galaxy.importer.models import BaseCollectionInfo, GalaxyCollectionInfo
+from galaxy.main import models
+
+
+
+class CollectionViewTests(TestCase):
+
+    def setUp(self):
+        self.namespace = models.Namespace.objects.create(
+            name='mynamespace',
+        )
+        self.collection = models.Collection.objects.create(
+            namespace=self.namespace,
+            name='mycollection',
+        )
+        self.version = models.CollectionVersion.objects.create(
+            collection=self.collection,
+            version='1.0.0',
+            contents='{}',
+        )
+
+    def test_get_collections(self):
+        client = APIClient()
+        resp = client.get("/api/internal/ui/collections/")
+        results = resp.json()['results']
+
+        collection = results[0]
+
+        assert collection['namespace'] == self.namespace.id
+        assert collection['name'] == self.collection.name
+        assert collection['latest_version']['version'] == self.version.version
--- a/galaxy/settings/local.py
+++ b/galaxy/settings/local.py
@@ -1,0 +1,10 @@
+from .testing import *
+
+
+DATABASES = {
+    'default': {
+        'NAME': 'galaxy',
+        'CONN_MAX_AGE': None,
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
+    }
+}
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,25 +36,12 @@ galaxy =
     templates/*.html
     templates/**/*.html
 
-[tool:pytest]
-# NOTE: This way won't work because env vars are already defined by compose
-# DJANGO_SETTINGS_MODULE=..
-# So instead of pytest-django we use pytest-env to override envvars forcefully
-env =
-    DJANGO_SETTINGS_MODULE=galaxy.settings.testing
-addopts =
-    --cov galaxy
-    --cov-report xml
-    --cov-report term
-    --cov-report html
-    --cov-config setup.cfg
-
 [coverage:run]
 omit = */migrations/*
        */test_*.py
 branch = True
 
 [coverage:report]
-fail_under = 39
+fail_under = 40
 
 [flake8]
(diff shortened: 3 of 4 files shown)


galaxy/settings/local.py
```diff
@@ -0,0 +1,10 @@
+from .testing import *  # noqa: F401,F403
```
cutwater: Is this file included into commit intentionally?

galaxy/api/v2/tests/test_collection_view.py
```diff
@@ -0,0 +1,32 @@
+from django.test import TestCase
+from rest_framework.test import APIClient
+
+from galaxy.main import models
+
+
+class CollectionViewTests(TestCase):
```
cutwater: You can extend `rest_framework.test.APITestCase` class that is inherited from `django.test.TestCase` and includes APIClient instance during setup.

galaxy/api/v2/tests/test_collection_view.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def setUp(self):
```
cutwater: parent class setUp() method call is missing

galaxy/api/v2/tests/test_collection_view.py
```diff
@@ -0,0 +1,32 @@
...
+        self.namespace = models.Namespace.objects.create(
```
cutwater: `setUp()` method is called before executing every test in test case class.

Testing database is not recreated for each test case, therefore creating namespace with non random name will cause conflict if more than one test in class or with other tests.

galaxy/api/v2/tests/test_collection_view.py
```diff
@@ -0,0 +1,32 @@
...
+            name='mynamespace',
+        )
```
cutwater: Probably better to cleanup created namespaces after test is finished.

`self.addCleanup(self.namespace.delete)`

setup.cfg
```diff
@@ -36,25 +36,12 @@ galaxy =
     templates/*.html
     templates/**/*.html
 
-[tool:pytest]
```
cutwater: Why these settings were moved outside configuration file?

galaxy/api/v2/tests/test_collection_view.py
ironfroggy: In fact, the main value point of `django.test.TestCase`, which `APITesetCase` inherits, is that each test is isolated from one another by virtue of being run in a separate transaction, which is rolled back at the end of the test. So, no need for cleanup and the created objects are already removed.

setup.cfg
cutwater: UPD: Moving coverage options seems reasonable to me, but why to move `DJANGO_SETTINGS_MODULE`?

galaxy/api/v2/tests/test_collection_view.py
ironfroggy: I can add that a `super().setUp()` here, but its not conventionally used in `setUp()` methods as they are explicitly hooks for the child class the parent does not provide any default behavior for, by definition.

galaxy/api/v2/tests/test_collection_view.py
ironfroggy: As stated, this is not necessary. All fixtures are already cleaned up between tests.

setup.cfg
ironfroggy: I moved this to facilitate running subsets of the full test suite, which is essential for swift workflows and also for `pytest` flags like `--ll` and `--lf`. In these cases the coverage generated is totally invalid, having been generated by a test run that didn't include all the relevant tests. So, I moved the flags from the default configuration to the invocation, in the Makefile, so they could differ based on use.

galaxy/settings/local.py
ironfroggy: local settings shouldn't get checked in, no. Thanks for the catch.

galaxy/api/v2/tests/test_collection_view.py
cutwater: It's true in case of django TestCase hierarchy, but for instance DRF and django base classes override `setUpClass()` method, so IMO it's good practice to call parent class `setUp()` in case there is any logic present.
Also it might be useful for generic parent classes that provide common mocks for example.

galaxy/api/v2/tests/test_collection_view.py
cutwater: I re-checked Django tests behavior and you're right, no need to cleanup database resources between tests.

galaxy/api/v2/tests/test_collection_view.py
cutwater: Ok

setup.cfg
cutwater: Ok, thanks for explanation.

PR #1718, opened 2019-04-09
Render collection-level readme on import

### LINKED ISSUE #1695: Save rendered collection-level readme on import
### Use Case
Allow for display of collection-level readme

### Proposed Solution
Save collection-level readme as rendered html for each import of collection version. Initial file format support will be markdown.

awcrosby:
Resolves #1695


galaxy/importer/collection.py
```diff
@@ -98,31 +98,39 @@ def _load_collection_manifest(self):
             self.collection_info = meta.collection_info
 
     def _load_collection_readme(self):
+        ALLOWED_TYPES = ['text/markdown', 'text/x-rst']
```
cutwater: Please move constant out of function

galaxy/importer/tests/test_readme_utils.py
```diff
@@ -0,0 +1,89 @@
...
+> NOTE: Tool 'feature' is _beta_
+
+* Item1
+* Item2
+'''
+
+
+class ReadmeRenderHtmlTestCase(TestCase):
+    def setUp(self):
+        self.Data = collections.namedtuple('Data', 'text mimetype')
```
cutwater: Why is this namedtuple part of `setUp()`?

galaxy/importer/tests/test_readme_utils.py
```diff
@@ -0,0 +1,89 @@
...
+    package_version: "1.2.0"
+
+> NOTE: Tool 'feature' is _beta_
+
+* Item1
+* Item2
+'''
+
+
+class ReadmeRenderHtmlTestCase(TestCase):
```
cutwater: It doesn't seem you're using any function provided by django test case here.
Why not use `unittest.TestCase`?

galaxy/importer/tests/test_readme_utils.py
```diff
@@ -0,0 +1,89 @@
...
+    def setUp(self):
+        self.Data = collections.namedtuple('Data', 'text mimetype')
+
+    def call_render(self, raw_text, mimetype):
+        readme_file = self.Data(text=raw_text, mimetype=mimetype)
+        return utils.readme.render_html(readme_file)
+
+    def test_render_simple(self):
+        html = self.call_render(TEXT_SIMPLE, 'text/x-rst')
+        self.assertEqual(html, '')
```
cutwater: Since galaxy uses `pytest` you can use bare `assert` operator for simple assertions, which is preferred. Example:

```
assert html == ''
assert '<script>' not in html
...
```

galaxy/importer/tests/test_readme_utils.py
awcrosby: Yes much better at the root of document

galaxy/importer/tests/test_readme_utils.py
awcrosby: Yes the base class `unittest.TestCase` would do fine, and with pytest `unittest` is not needed, so this was simplified into function-based tests.

Issue #1609, opened 2019-03-01
Refactor My Imports


newswangerd:
- [x] migrate to react
- [x] only display the user's imports
- [x] admins should be able to see all imports (but not regular users)
- [x] users with multiple namespaces should see all their namespaces
- [x] add extra info for troubleshooting

PR #1743, opened 2019-04-16
Enable collections on My Imports

newswangerd:
Changes:
- My imports are now routed via `/my-imports/<namespace-id>/` and accepts page, page_size, name and type as query params.
  - Visiting `/my-imports/` will load whichever namespace matches your username as the default namespace
  - Other namespaces that the user has access to are available via a dropdown menu.
- Rewrite my imports using react
- Import collections from My Content redirects the user to My Imports

image

fixes: #1609


galaxy/api/internal/views/imports.py
```diff
@@ -97,7 +97,7 @@ def get_repsository_queryset(self):
 
         name = self.request.query_params.get('name')
         if name:
-            qs = qs.filter(repository__name__iexact=name)
+            qs = qs.filter(repository__name__icontains=name)
```
ironfroggy: This can change the result from a single collection to multiple, which doesn't seem right.

newswangerd: The problem is that we use that we need to use that query parameter for the filter which means that `iexact` doesn't return anything when you type a partial match into the filter box. I thought that was a really frustrating user experience so I switched it to `icontains`. If we want to guarantee that we are only getting one repo or collection back, I think we should probably add an ID filter as well.

Issue #1729, opened 2019-04-10
Collection version detail API endpoint


awcrosby:
Part of #1661
Display collection version info based on namespace, name, and specific version such as “1.4.2”
`GET /api/v2/collections/<namespace>/<name>/versions/<version>/`
`GET /api/v2/collection-versions/<id>/`

Issue #1728, opened 2019-04-10
Collection version list API endpoint


awcrosby:
Part of #1661
Display a list of versions for a collection, sorted by semantic_version: 
`GET /api/v2/collections/<namespace>/<name>/versions/`
`GET /api/v2/collections/<id>/versions/`

Issue #1727, opened 2019-04-10
Collection detail API endpoint


awcrosby:
Part of #1661
Display collection info based on namespace and name
`GET /api/v2/collections/<namespace>/<name>/`
`GET /api/v2/collections/<id>/`

Issue #1742, opened 2019-04-15
Errors returned for POST /api/v2/collections seem inconsistent


alikins:
##### SUMMARY

The way errors are returned from POST /api/v2/collections feels odd... 
Success returns JSON like

``` json
{"task":"/api/v2/collection-imports/4/"}
```
But the JSON returned on errors varies.

Uploading a collection using a filename indicating a non existent namespace ('blipalikins') returns a '400 - Bad Request" http status and a body with list with one error message in it:

``` json
["Namespace \"blipalikins\" does not exist."]
```

Using a filename with an invalid version string format returns a 
'400 - Bad Request" http status and a body with a dictionary with a "non_field_errors" with a list of error messages as value:

``` json
{"non_field_errors":["Invalid version string: '0.1.175.4'"]}
```

Attempting to publish a collection that has already been publish before results in 
a '409 Conflict' http status and returning a dictionary with a "detail" key and a string value:

``` json
{"detail":"Collection already exists."}
```

The different types and structures returned makes it difficult to present the errors to the user in the client code.

Galaxy is devel branch 3.0.3.dev676+g770810d
770810d20940a3ef94f46da16e653a04748f90a3

##### EXPECTED RESULTS

For error cases, the json body could be more standardized with a consistent structure.

For a strawman example, something like:

``` json
{
   "errors": ["The collection artifact filename 'foo-bar-1.2.3.4' has a version number that is not a valid semantic version format: '1.2.3.4'"]
}
```

##### ACTUAL RESULTS

I don't see server side tracebacks for these cases. Haven't verified but would guess default exception handlers are creating the response bodies from different exception types.

alikins:
(some notes from slack I'll forget otherwise)

> That particular endpoint has characteristics that tend to make for unusual error handling (ie, things like multipart form support, it being kind of "async", the token auth and authz tied to it, the fuzzy line between parsing/validating the uploaded body and validating the artifact and the content in the collection. Task queuing in general tends to involve inconvenient error paths...)

ironfroggy:
Sorry for the mistaken close/open i was writing a comment, deciding against it, and in my pre-coffee brain state thought i was hitting a cancel button 😬

PR #1745, opened 2019-04-16
Add collection and version api/v2 GET endpoints

awcrosby:
Add api/v2 GET views, with multiple url endpoint formats for:
- Collection detail #1727
- CollectionVersion list #1728 
- CollectionVersion detail #1729

### DIFF
--- a/galaxy/api/v2/pagination.py
+++ b/galaxy/api/v2/pagination.py
@@ -18,7 +18,7 @@
 from rest_framework import pagination
 
 
-class CustomPagination(pagination.PageNumberPagination):
+class DefaultPagination(pagination.PageNumberPagination):
     page_size = 10
     page_size_query_param = 'page_size'
-    max_page_size = 1000
+    max_page_size = 100
--- a/galaxy/api/v2/pagination.py
+++ b/galaxy/api/v2/pagination.py
@@ -1,0 +1,54 @@
+# (c) 2012-2019, Ansible by Red Hat
+#
+# This file is part of Ansible Galaxy
+#
+# Ansible Galaxy is free software: you can redistribute it and/or modify
+# it under the terms of the Apache License as published by
+# the Apache Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# Ansible Galaxy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# Apache License for more details.
+#
+# You should have received a copy of the Apache License
+# along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+from rest_framework import pagination
+from rest_framework.utils.urls import remove_query_param, replace_query_param
+
+
+class CustomPagination(pagination.PageNumberPagination):
+    page_size = 10
+    page_size_query_param = 'page_size'
+    max_page_size = 1000
+
+    # overrides to create relative URL
+    def get_next_link(self):
+        if not self.page.has_next():
+            return None
+        url = self.request.get_full_path()
+        page_number = self.page.next_page_number()
+        return replace_query_param(url, self.page_query_param, page_number)
+
+    # overrides to create relative URL
+    def get_previous_link(self):
+        if not self.page.has_previous():
+            return None
+        url = self.request.get_full_path()
+        page_number = self.page.previous_page_number()
+        if page_number == 1:
+            return remove_query_param(url, self.page_query_param)
+        return replace_query_param(url, self.page_query_param, page_number)
+
+    # TODO: decide - keep if want to rename "next" to "next_link", etc.
+    # def get_paginated_response(self, data):
+    #     next_link = self.get_next_link()
+    #     previous_link = self.get_previous_link()
+    #     return response.Response(OrderedDict([
+    #         ('count', self.page.paginator.count),
+    #         ('next_link', next_link),
+    #         ('previous_link', previous_link),
+    #         ('results', data),
+    #     ]))
--- a/galaxy/api/v2/pagination.py
+++ b/galaxy/api/v2/pagination.py
@@ -16,39 +16,9 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 from rest_framework import pagination
-from rest_framework.utils.urls import remove_query_param, replace_query_param
 
 
 class CustomPagination(pagination.PageNumberPagination):
     page_size = 10
     page_size_query_param = 'page_size'
     max_page_size = 1000
-
-    # overrides to create relative URL
-    def get_next_link(self):
-        if not self.page.has_next():
-            return None
-        url = self.request.get_full_path()
-        page_number = self.page.next_page_number()
-        return replace_query_param(url, self.page_query_param, page_number)
-
-    # overrides to create relative URL
-    def get_previous_link(self):
-        if not self.page.has_previous():
-            return None
-        url = self.request.get_full_path()
-        page_number = self.page.previous_page_number()
-        if page_number == 1:
-            return remove_query_param(url, self.page_query_param)
-        return replace_query_param(url, self.page_query_param, page_number)
-
-    # TODO: decide - keep if want to rename "next" to "next_link", etc.
-    # def get_paginated_response(self, data):
-    #     next_link = self.get_next_link()
-    #     previous_link = self.get_previous_link()
-    #     return response.Response(OrderedDict([
-    #         ('count', self.page.paginator.count),
-    #         ('next_link', next_link),
-    #         ('previous_link', previous_link),
-    #         ('results', data),
-    #     ]))
--- a/galaxy/api/v2/serializers/collection.py
+++ b/galaxy/api/v2/serializers/collection.py
@@ -28,7 +28,119 @@ from .tasks import BaseTaskSerializer
 __all__ = (
     'CollectionImportSerializer',
     'CollectionUploadSerializer',
+    'CollectionSerializer',
+    'VersionSummarySerializer',
+    'VersionDetailSerializer',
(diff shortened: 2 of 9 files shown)


galaxy/api/v2/serializers/collection.py
```diff
@@ -28,9 +28,121 @@
 __all__ = (
     'CollectionImportSerializer',
     'CollectionUploadSerializer',
+    'CollectionSerializer',
+    'VersionSummarySerializer',
+    'VersionDetailSerializer',
 )
 
 
+class NamespaceObjectField(serializers.Field):
```
awcrosby: considering moving these custom fields to `api/v2/fields.py`

galaxy/api/v2/views/collection_version.py
```diff
@@ -15,21 +15,70 @@
...
+
+    if pk:
+        return get_object_or_404(models.Collection, pk=pk)
+
+    ns = get_object_or_404(models.Namespace, name=ns_name)
+    return get_object_or_404(models.Collection, namespace=ns, name=name)
+
+
+class VersionListView(generics.ListAPIView):
     permission_classes = (IsAuthenticated, )
```
alikins: Guessing this is the cause of the auth check / 401 mentioned below.

Does VersionListView / CollectionListView need auth?

alikins:
With the 'isAuthenticated' checks removed from the Version views in my local dev galaxy, mazer can find the collection and best version and be prepared to download with the branch at https://github.com/ansible/mazer/pull/226

alikins:
Whats necessary to get the /api/v2 stuff exposed in the /api html view? (http://localhost:8000/api/ in  a browswer)

galaxy/api/v2/serializers/collection.py
```diff
@@ -65,7 +177,7 @@ def get_imported_version(self, obj):
             return None
         return {
             'id': obj.imported_version.pk,
-            'href': reverse('api:v2:collection-version-detail',
+            'href': reverse('api:v2:version-detail-by-id',
```
cutwater: AFAIK `by-id` suffix is not necessary, you may have multiple routes with same name, but different parameters.

galaxy/api/v2/views/collection.py
```diff
@@ -36,10 +37,34 @@
...
     'CollectionListView',
+    'CollectionDetailView',
 ]
 
 
 # FIXME(cutwater): Implement consistent error reporting format.
+
+
+class CollectionDetailView(views.APIView):
+    permission_classes = (IsAuthenticated, )
```
cutwater: Shouldn't collection info be available to anonymous users?

galaxy/api/v2/views/collection.py
```diff
@@ -36,10 +37,34 @@
...
+
+    def get(self, request, *args, **kwargs):
+        """Return a collection."""
+        collection = self._get_collection(kwargs)
+        serializer = serializers.CollectionSerializer(collection)
+        return Response(serializer.data)
+
+    def _get_collection(self, kwargs):
```
cutwater: Request arguments are available as `self.kwargs`

galaxy/api/v2/views/collection_version.py
```diff
@@ -15,21 +15,70 @@
...
 __all__ = (
-    'CollectionVersionView',
+    'VersionListView',
+    'VersionDetailView',
 )
 
 
-class CollectionVersionView(views.APIView):
+# TODO(awcrosby): Move to views/utils.py and reuse across views
+def _lookup_collection(kwargs):
```
cutwater: ditto

alikins:
I did a rudimentary merge of this branch (as of 002c38d269b701a3f5b4975e3107bdd68b4196e1) and #1746 at https://github.com/ansible/galaxy/compare/devel...alikins:col_list_col_artifact

Using it with https://github.com/ansible/galaxy/issues/1742  seems to work

galaxy/api/v2/views/collection_version.py
```diff
@@ -15,21 +15,70 @@
...
+class VersionListView(generics.ListAPIView):
     permission_classes = (IsAuthenticated, )
+    serializer_class = serializers.VersionSummarySerializer
+    pagination_class = CustomPagination
+
+    def get_queryset(self):
+        """Return list of versions for a specific collection."""
+        collection = _lookup_collection(self.kwargs)
+        return models.CollectionVersion.objects.filter(collection=collection)
+
```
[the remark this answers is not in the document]
awcrosby: Yeah I still have the semver sorting to do on version list view

galaxy/api/v2/views/collection_version.py
awcrosby: I pulled the permission_class from the POST, I will update these GET views to the AllowAny permission

awcrosby:
> Whats necessary to get the /api/v2 stuff exposed in the /api html view? (http://localhost:8000/api/ in a browswer)

Looks like need to add ApiV2RootView like this https://github.com/ansible/galaxy/blob/devel/galaxy/api/views/views.py#L132

galaxy/api/v2/views/collection_version.py
```diff
@@ -15,21 +15,70 @@
...
+class VersionListView(generics.ListAPIView):
+    permission_classes = (AllowAny, )
+    serializer_class = serializers.VersionSummarySerializer
+    pagination_class = CustomPagination
+
+    def get_queryset(self):
+        """Return list of versions for a specific collection."""
+        collection = _lookup_collection(self.kwargs)
+        return models.CollectionVersion.objects.filter(collection=collection)
```
awcrosby: @cutwater @alikins To sort a collection version list by semver, we need to evaluate the queryset, but for the drf pagination classes, it takes a queryset (also evaluating a large queryset is not ideal). An option was mentioned earlier to change `CollectionVersion.version` type from `CharField` to `semantic_version.django_fields.VersionField`, this looks like a good solution, thoughts?

cutwater: 1. Ok, we may replace `CharField` with `VersionField`, it will make it easier to work with, but doesn't really solve large query evaluation problem, because `VersionField` is just a wrapper for conversion between python code and database.
2. DRF pagination classes able to paginate data or queryset.

galaxy/api/v2/pagination.py
```diff
@@ -0,0 +1,54 @@
...
+# Apache License for more details.
+#
+# You should have received a copy of the Apache License
+# along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+from rest_framework import pagination
+from rest_framework.utils.urls import remove_query_param, replace_query_param
+
+
+class CustomPagination(pagination.PageNumberPagination):
```
cutwater: Why did you implement custom pagination class?

awcrosby: api/v1 gives multiple next and previous links with a hardcoded removal of ‘api/v1’:
```
"next": "/repositories/?page=2",
"next_link": "/api/v1/repositories/?page=2",
"previous": null,
"previous_link": null,
```
api/v2 should be more standard unless otherwise needed:
```
"next": "/api/v2/collections/1/versions/?page=2",
"previous": null,
```

This v2 pagination class can be moved to the global default, and v1 endpoints changed to use the old pagination class - but should be done in separate PR

alikins:
What is the current plan for where the download or artifact download url will be provided?

Will the download_url field of the VersionDetailView / 'collections/<str:namespace>/<str:name>/versions/<str:version>/' return include it?
Elsewhere?

Will  'collections/<str:namespace>/<str:name>/versions/<str:version>/' result include the full url or relative path? Assuming the download url will potentially live on a CDN, full url would be better.

awcrosby:
> What is the current plan for where the download or artifact download url will be provided?
> 
> Will the download_url field of the VersionDetailView / 'collections/str:namespace/str:name/versions/str:version/' return include it?

@alikins `download_url` field of VersionDetailView will be `"/api/v2/collections/str:namespace/str:name/versions/str:version/artifact/"`

> Will 'collections/str:namespace/str:name/versions/str:version/' result include the full url or relative path? Assuming the download url will potentially live on a CDN, full url would be better.

I am assuming the api will not resolve the above `".../artifact/"` url to what it will redirect to

galaxy/api/v2/pagination.py
awcrosby: Custom pagination class logic was narrowed to just custom page size

galaxy/api/v2/views/collection_version.py
awcrosby: DRF is now paginating the data, and versions are sorted by semantic version

galaxy/api/v2/views/collection_version.py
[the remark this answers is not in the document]
awcrosby: Versions are now sorted by semantic version

galaxy/api/v2/pagination.py
```diff
@@ -0,0 +1,24 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# Apache License for more details.
+#
+# You should have received a copy of the Apache License
+# along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+from rest_framework import pagination
+
+
+class CustomPagination(pagination.PageNumberPagination):
```
cutwater: _(optional)_ `PageNumberPagination` or `DefaultPagination` name in context of project may be better.

galaxy/api/v2/pagination.py
```diff
@@ -0,0 +1,24 @@
...
+    page_size = 10
+    page_size_query_param = 'page_size'
+    max_page_size = 1000
```
cutwater: Please reduce it to reasonably lower level

awcrosby: Agreed (note, we have this large max_page_size in v1/api, but that will be deprecated). Lowered from 1000 to 100.

PR #1753, opened 2019-04-17
Avoid combining bash and Makefile variable syntax in the same stateme…

ironfroggy:
…nt: confusing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,8 +210,14 @@ dev/test:
 dev/test/changed:
 	@echo "Running changed tests"
 	@$(DOCKER_COMPOSE) exec galaxy bash -c '\
-		export DJANGO_SETTINGS_MODULE=galaxy.settings.testing \
-		git status | grep 'test' | xargs pytest -s -vvv --reuse-db \
+		source $(GALAXY_VENV)/bin/activate; \
+		export DJANGO_SETTINGS_MODULE="galaxy.settings.testing"; \
+		git status | grep test | cut -d: -f2 > /tmp/tests_changed ;\
+		if [[ -s /tmp/tests_changed ]]; then \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -210,8 +210,14 @@ dev/test:
...
-		export DJANGO_SETTINGS_MODULE=galaxy.settings.testing \
-		git status | grep 'test' | xargs pytest -s -vvv --reuse-db \
+		source $(GALAXY_VENV)/bin/activate; \
+		export DJANGO_SETTINGS_MODULE="galaxy.settings.testing"; \
+		git status | grep test | cut -d: -f2 > /tmp/tests_changed ;\
+		if [[ -s /tmp/tests_changed ]]; then \
+				cat /tmp/tests_changed | xargs pytest -s -vvv --reuse-db \
+		;else \
+				echo "No test changes found. make dev/test to run all tests." \
+		;fi \
```
awcrosby: edited the syntax a bit, then it seemed to work:
```
export DJANGO_SETTINGS_MODULE=galaxy.settings.testing; \
git status | grep "test" | xargs pytest -s -vvv --reuse-db; \
source $(GALAXY_VENV)/bin/activate; \
export DJANGO_SETTINGS_MODULE="galaxy.settings.testing"; \
git status | grep test | cut -d: -f2 > /tmp/tests_changed ;\
if [[ -s /tmp/tests_changed ]]; then \
		cat /tmp/tests_changed | xargs pytest -s -vvv --reuse-db \
;else \
		echo "No test changes found. make dev/test to run all tests." \
;fi \
'
```

I see there is a lot more output than `dev/test` such as apply migrations, and `Adding permission 'Permission object (294)'` is all that expected?

PR #1758, opened 2019-04-18
Add API endpoint for getting paginated lists of collections and repositories

newswangerd:


galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,101 @@
...
+
+        start = (page * page_size) - page_size
+        end = page * page_size
+
+        collection_results = collections[start:end]
+
+        # If the collections fill the whole page, don't bother loading the
+        # repos
+        if (len(collection_results) == page_size):
+            repo_results = []
```
ironfroggy: it may be more correct here to say `models.Repository.objects.none()`

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,101 @@
...
+        else:
+            # Django gets cranky if you try to slice a queryset with a negative
+            # index
+            r_start = start - collection_count
+            r_start = 0 if r_start < 0 else r_start
```
ironfroggy: `r_start = max(0, start - collection_count)`

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class CombinedList(views.APIView):
+    def get(self, request):
+        # params: page, page size, namespace, name, description, type
+        # order by: name, downloads
+        namespace = request.GET['namespace']
+        page = int(request.GET.get('page', 1))
+        page_size = int(request.GET.get('page_size', 10))
+        type = request.GET.get('type', None)
```
alikins: I would avoid use of 'type' as a local variable name if possible. It's kind of vague, but more importantly since it's also defined globally, it's very easy to accidentally not define it locally (since it _is_ defined) and end up getting weird behavior when comparing the identity of `<class 'type'>` with a string. 

Maybe 'package_type'? ( avoiding 'content_type' since that has another meaning already). Or 'type_' or '_type' perhaps.

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,101 @@
...
+        # repos
+        if (len(collection_results) == page_size):
+            repo_results = []
+        else:
+            # Django gets cranky if you try to slice a queryset with a negative
+            # index
+            r_start = start - collection_count
+            r_start = 0 if r_start < 0 else r_start
+
+            repo_results = repos[r_start:end-collection_count]
```
ironfroggy: I think you actually want `len(collection_results)` rather than `collection_count`, because you want to offset the repository results by the number of collections in this page, not the total number of collections.

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,101 @@
...
+#
+# You should have received a copy of the Apache License
+from rest_framework import views
+from galaxy.main import models
+from rest_framework import response
+from galaxy.api.internal import serializers as internal_serializers
+from galaxy.api import serializers as v1_serializers
+
+
+class CombinedList(views.APIView):
```
alikins: For an internal only API, if this view is only going to be collections and repositories, it might be more explicit to just call it something like ReposAndCollectionsList().  Assuming it's not going to gain support for another type of <whatever we call repositories and collections> that is. Not a big deal either way for an internal only api though.

galaxy/api/internal/views/combined.py
newswangerd: I think my approach is correct because in this case `end` is the index of the last item in the list if the list was a combined list of repos and collections so we need to offset the starting point for repos by the total number of collections

Lets say we have 21 collections and 40 repos. On page 3 (with page size 10) `end=30` and  we would expect to see `collection_results=collections[20:21]` and `repo_results=repos[0:9]`. If we used `len(collection_results)` then the function would return `repos[0:30-1]`, which doesn't work.

galaxy/api/internal/views/combined.py
cutwater: `_` suffix is normally used to avoid name collision with python built-ins. so `type_` is preferred.

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,108 @@
...
+                {
+                    'count': repo_count,
+                    'results': v1_serializers.RepositorySerializer(
+                        repo_results,
+                        many=True,
+                    ).data,
+                }
+        }
+
+        return(response.Response(result))
```
cutwater: Parentheses are not required here.

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,108 @@
...
+# the Apache Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# Ansible Galaxy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# Apache License for more details.
+#
+# You should have received a copy of the Apache License
+from rest_framework import views
```
cutwater: Please group imports according to PEP8.

> Imports should be grouped in the following order:
>
> Standard library imports.
> Related third party imports.
> Local application/library specific imports.

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,108 @@
...
+from galaxy.main import models
+from rest_framework import response
+from galaxy.api.internal import serializers as internal_serializers
+from galaxy.api import serializers as v1_serializers
+from rest_framework import status
+
+
+class RepoAndCollectionList(views.APIView):
+    def get(self, request):
+        namespace = request.GET.get('namespace', None)
```
cutwater: It's ok to use `request.GET`, but DRF provides a more correctly named alias `request.query_params`. It's preferred over the standard django `request.GET`.
https://www.django-rest-framework.org/api-guide/requests/#query_params

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,108 @@
...
+        page = int(request.GET.get('page', 1))
+        page_size = int(request.GET.get('page_size', 10))
+        package_type = request.GET.get('type', None)
+        order = request.GET.get('order', 'name')
+
+        if not namespace:
+            return response.Response(
```
cutwater: You should use `APIException` or one of its derived classes.

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,108 @@
...
+                {'detail': 'The namespace parameter is required'},
+                status=status.HTTP_400_BAD_REQUEST
+            )
+
+        collection_filter = {
```
cutwater: filter**s** maybe?

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,108 @@
...
+            collection_count = collections.count()
+
+        start = (page * page_size) - page_size
+        end = page * page_size
+
+        collection_results = collections[start:end]
+
+        # If the collections fill the whole page, don't bother loading the
+        # repos
+        if (len(collection_results) == page_size):
```
cutwater: Parentheses are not required.

galaxy/api/internal/views/combined.py
cutwater: Also you may use DRF serializer to parse arguments.

galaxy/api/internal/views/combined.py
newswangerd: omg... I'm turning into a Javascript developer...

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,111 @@
...
+            'provider_namespace__namespace__name': namespace
+        }
+
+        if 'name' in request.GET:
+            collection_filters['name__icontains'] = request.GET['name']
+            repo_filters['name__icontains'] = request.GET['name']
+
+        if 'description' in request.GET:
+            collection_filters[
+                'latest_version__metadata__description__icontaines'
```
ironfroggy: typo in `icontaines`

galaxy/api/internal/views/combined.py
```diff
@@ -0,0 +1,111 @@
...
+            repos = models.Repository.objects.filter(
+                **repo_filters).order_by(order)
+            repo_count = repos.count()
+
+        if package_type == 'repository':
+            collections = models.Collection.objects.none()
+            collection_count = 0
+        else:
+            collections = models.Collection.objects.filter(
+                **collection_filters).order_by(order)
```
ironfroggy: invalid values for `order` should be handled and reported properly

galaxy/api/internal/views/combined.py
```diff
@@ -109,3 +109,10 @@ def get(self, request):
         }
 
         return response.Response(result)
+
+    def _validate_order(self, order):
+        allowed_orders = ['download_count', 'name', '-download_count', '-name']
+        if order not in allowed_orders:
+            raise InvalidParams(
+                detail="Order must be one of " + str(allowed_orders)
```
awcrosby: noticed sometimes the use of `detail=` vs `details=` in `InvalidParams()`.

also very minor... for future consistency, we usually use `%` or `.format()` for string concatenation (and I guess the `+` spacing looks a bit odd in an argument)

ironfroggy: According to documentation on the base exception class `APIException`, the correct parameter is `detail` so the other uses are incorrect and may well cause errors when they are hit. Two notes:

1. @newswangerd Let's make sure the original reviewers sign off on a PR before merging it when it has specifically been marked with requests. A little delay can save us some pain and breakage.
2. The lines using `details` will actually cause an error of their own (exceptions causing exceptions! the humanity of it all!) but all the tests are passing. That probably means there are some gaps in the tests we could fill in by making sure these failure cases are covered, not just the successful cases.

https://www.django-rest-framework.org/api-guide/exceptions/#apiexception

PR #1760, opened 2019-04-23
Implement custom API exception handler

cutwater:
* Implement custom error handler for API v2 and internal views.
* Implement base class `APIView` and generic classes to provide
  custom exception handler and avoid conflict with legacy v1
  configuration that relies on global settings.
* Add generic classes that are derived from custom base `APIView`.
* Update existing API classes to derive from custom base classes.

### DIFF
--- a/galaxy/api/base.py
+++ b/galaxy/api/base.py
@@ -1,0 +1,210 @@
+# (c) 2012-2019, Ansible by Red Hat
+#
+# This file is part of Ansible Galaxy
+#
+# Ansible Galaxy is free software: you can redistribute it and/or modify
+# it under the terms of the Apache License as published by
+# the Apache Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# Ansible Galaxy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# Apache License for more details.
+#
+# You should have received a copy of the Apache License
+# along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+from django.core import exceptions as dj_exc
+from django import http as dj_http
+from rest_framework import exceptions as drf_exc
+from rest_framework import generics
+from rest_framework import mixins
+from rest_framework.response import Response
+from rest_framework.settings import api_settings
+from rest_framework import status as http_codes
+from rest_framework import views
+
+
+def exception_handler(exc, context):
+    if isinstance(exc, dj_http.Http404):
+        data = {
+            'code': drf_exc.NotFound.default_code,
+            'message': drf_exc.NotFound.default_detail,
+        }
+        views.set_rollback()
+        return Response(data, status=http_codes.HTTP_404_NOT_FOUND)
+
+    if isinstance(exc, dj_exc.PermissionDenied):
+        data = {
+            'code': drf_exc.PermissionDenied.default_code,
+            'message': drf_exc.PermissionDenied.default_detail,
+        }
+        views.set_rollback()
+        return Response(data, status=http_codes.HTTP_403_FORBIDDEN)
+
+    if isinstance(exc, drf_exc.APIException):
+        headers = {}
+        if getattr(exc, 'auth_header', None):
+            headers['WWW-Authenticate'] = exc.auth_header
+        if getattr(exc, 'wait', None):
+            headers['Retry-After'] = '%d' % exc.wait
+
+        data = {
+            'code': exc.default_code,
+            'message': exc.default_detail,
+        }
+        errors = []
+
+        if isinstance(exc.detail, list):
+            errors = [{'code': e.code, 'message': e} for e in exc.detail]
+        elif isinstance(exc.detail, dict):
+            for field, messages in exc.detail.items():
+                if isinstance(messages, str):
+                    messages = [messages]
+                for message in messages:
+                    error = {'code': message.code, 'message': message}
+                    if field != api_settings.NON_FIELD_ERRORS_KEY:
+                        error['field'] = field
+                    errors.append(error)
+        else:
+            data['code'] = exc.detail.code
+            data['message'] = exc.detail
+
+        if len(errors) == 1 and 'field' not in errors[0]:
+            data.update(errors[0])
+        elif errors:
+            data['errors'] = errors
+
+        views.set_rollback()
+        return Response(data, status=exc.status_code, headers=headers)
+
+    return None
+
+
+class ExceptionHandlerMixin:
+
+    def get_exception_handler(self):
+        return exception_handler
+
+
+class APIView(ExceptionHandlerMixin,
+              views.APIView):
+    """Base class for API views."""
+    pass
+
+
+class GenericAPIView(APIView,
+                     generics.GenericAPIView):
+    """Base class for generic API views."""
+    pass
+
+
+class CreateAPIView(mixins.CreateModelMixin,
+                    GenericAPIView):
+    """
+    Concrete view for creating a model instance.
+    """
+    def post(self, request, *args, **kwargs):
+        return self.create(request, *args, **kwargs)
+
+
+class ListAPIView(mixins.ListModelMixin,
+                  GenericAPIView):
+    """
+    Concrete view for listing a queryset.
+    """
+    def get(self, request, *args, **kwargs):
+        return self.list(request, *args, **kwargs)
+
+
+class RetrieveAPIView(mixins.RetrieveModelMixin,
+                      GenericAPIView):
+    """
+    Concrete view for retrieving a model instance.
+    """
+    def get(self, request, *args, **kwargs):
+        return self.retrieve(request, *args, **kwargs)
+
+
+class DestroyAPIView(mixins.DestroyModelMixin,
+                     GenericAPIView):
+    """
+    Concrete view for deleting a model instance.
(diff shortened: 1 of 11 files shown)


galaxy/api/base.py
```diff
@@ -0,0 +1,204 @@
...
+class ExceptionHandlerMixin:
+
+    def get_exception_handler(self):
+        return exception_handler
+
+
+class APIView(ExceptionHandlerMixin,
+              views.APIView):
+    """Base class for API views."""
+    pass
```
ironfroggy: `pass` is not necessary because the docstring already provides a body

galaxy/api/tests/test_exception_handler.py
```diff
@@ -0,0 +1,67 @@
...
+
+        data = response.data
+        errors = sorted(data['errors'], key=operator.itemgetter('message'))
+        assert data['message'] == 'Invalid input.'
+        assert errors == [
+            {'message': 'First bar message.', 'field': 'bar'},
+            {'message': 'Foo error message.', 'field': 'foo'},
+            {'message': 'Other message.'},
+            {'message': 'Second bar message.', 'field': 'bar'},
+        ]
```
ironfroggy: The order here is only to help with testing, but does not assert any particular ordering in the errors list that comes in the response, right?

galaxy/api/base.py
cutwater: Indeed, but it's common approach and I would like to leave it for better readability.

galaxy/api/tests/test_exception_handler.py
cutwater: That's true.

galaxy/api/v2/views/collection.py
```diff
@@ -110,7 +110,9 @@ def _check_version_conflict(self, namespace, filename):
         except dj_exc.ObjectDoesNotExist:
             pass
         else:
-            raise CollectionExistsError()
+            raise CollectionExistsError(
+                f'Collection "{filename.namespace}-{filename.name}'
+                f'-{filename.version}" already exists.')
```
awcrosby: Should we put this extra detail in the `_save_artifact()` `CollectionExistsError()` also?

galaxy/api/base.py
```diff
@@ -0,0 +1,204 @@
...
+            headers['WWW-Authenticate'] = exc.auth_header
+        if getattr(exc, 'wait', None):
+            headers['Retry-After'] = '%d' % exc.wait
+
+        data = {'message': exc.default_detail}
+        errors = []
+
+        if isinstance(exc.detail, list):
+            errors = [{'message': e} for e in exc.detail]
+        elif isinstance(exc.detail, dict):
```
alikins: Can exc.detail be anything other than a list?  https://www.django-rest-framework.org/api-guide/exceptions/#inspecting-api-exceptions claims:

> .detail - Return the textual description of the error.

The django_rest_framework.exceptions.APIException subclasses all seem to set the `detail` attribute to text.

alikins: Looks like drf.exceptions._get_exception_details() can return a list, .detail can be other things that text.

galaxy/api/base.py
```diff
@@ -0,0 +1,204 @@
...
+            for field, messages in exc.detail.items():
+                if isinstance(messages, str):
+                    messages = [messages]
+                for message in messages:
+                    error = {'message': message}
+                    if field != api_settings.NON_FIELD_ERRORS_KEY:
+                        error['field'] = field
+                    errors.append(error)
+        else:
+            data['message'] = exc.detail
```
alikins: Worthwhile to also include a `data['detail'] = exc.detail` always?[1]  So the stuff that does have structured details, we always return that as well? Clients like mazer wouldn't use `data['detail']` directly (preferring the `data['messages']` that we have ensured are in a consistent format) but may be useful other tools or troubleshooting.

[1] Assuming exc.detail is serializable

galaxy/api/base.py
```diff
@@ -0,0 +1,204 @@
...
+
+        if len(errors) > 1:
+            data['errors'] = errors
+        elif len(errors) == 1:
+            data.update(errors[0])
+
+        views.set_rollback()
+        return Response(data, status=exc.status_code, headers=headers)
+
+    return None
```
alikins: Worthwhile extracting the APIException() handler logic to it's own method? Would have to return `data` and `headers` and possible a `status_code`[1]. Something like:

``` python

if isinstance(exc, exceptions.APIException):
  data, status_code, headers = handle_api_exception(exc, context)
  views.set_rollback()
  return Response(data, status=status_code, headers=headers)
```


[1] if it ever was needed to return a different status_code in the Response than the one in exc.status_code

alikins:
(just a note related to the exception_handler() idea while I was thinking about it, not suggesting this needs to be changed...)

The api.base.exception_handler() is based on the same approach drf.views.exception_handler() uses which is probably the correct approach here.

One possible alternative construct is to build a try/except around `raise exc`. I've used that approach before and it has some interesting aspects. For an example of the construct:

``` python

def exception_handler(exc, context):
    try:
        raise exc
    except Http404 as e:
        log.exception(e)
        msg = 'Not found.'
        data = {'message': str(msg)}
    except PermissionDenied as e:
        log.exception(e)
        msg = 'Permission denied'
        data = {'message': str(msg)}
    except exceptions.ApiException as e:
        log.exception(e)
        data, status_code, headers = handle_api_exception(exc, context)
    finally:
        views.set_rollback()
        return Response(data, status=http_codes.HTTP_404_NOT_FOUND)
```

The main benefit is it looks like exception handling, and uses the exact same exception matching as typical (as opposed to similar if branching on isinstance()). And log.exception() and/or logging calls with exc_info/stack_info works as expected  [that may be okay in the isinstance() case as well, haven't tested...]

galaxy/api/tests/test_exception_handler.py
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+    def test_http_permission_denied(self):
+        error = PermissionDenied('Permission denied.')
+        response = exception_handler(error, None)
+        assert response.data == {
+            'message': 'Permission denied.',
+        }
+
+    def test_api_exc_string(self):
```
alikins: Maybe test cases for other detail types?  `ApiException(detail=None)`, `ApiException(detail={'some_key':'some_value'})`, `ApiException(detail=rest_framework.utils.serializer_helpers.ReturnList(serializer=SomeSerializer))`?

galaxy/api/base.py
cutwater: DRF documentation is incorrect. 

According to implementation of `APIException` class and standard `exception_handler`, the `.detail` attribute may be either `list` or its subclass, `dict` or its subclass and string (specifically ErrorDetail object, which inherits string and has `.code` attribute).

galaxy/api/base.py
cutwater: See my previous comment. All error details are normalized in DRF exception hierarchy. It can be either `list`, `dict`, `str` or their sub-classes.
All other types are normalized to string representation.
Indeed, there might be recursive structures and we may need to handle this case.

So in this case your proposed `detail` field would be no more than duplicate of `message`.
If we ever need structured details we may extend this API by adding new fields and using custom exception hierarchy handled appropriately in exception handler.

galaxy/api/base.py
cutwater: Only as a code decoupling. This is only one place where proposed function can be used.

galaxy/api/tests/test_exception_handler.py
cutwater: `ReturnList` and `ReturnDict` classes are subclasses of `list` and `dict` respectively and add additional reference to `serializer`, which is not uses during error serialization.

cutwater:
> (just a note related to the exception_handler() idea while I was thinking about it, not suggesting this needs to be changed...)
> 
> The api.base.exception_handler() is based on the same approach drf.views.exception_handler() uses which is probably the correct approach here.
> 
> One possible alternative construct is to build a try/except around `raise exc`. I've used that approach before and it has some interesting aspects. For an example of the construct:
> 
> ```python
> def exception_handler(exc, context):
>     try:
>         raise exc
>     except Http404 as e:
>         log.exception(e)
>         msg = 'Not found.'
>         data = {'message': str(msg)}
>     except PermissionDenied as e:
>         log.exception(e)
>         msg = 'Permission denied'
>         data = {'message': str(msg)}
>     except exceptions.ApiException as e:
>         log.exception(e)
>         data, status_code, headers = handle_api_exception(exc, context)
>     finally:
>         views.set_rollback()
>         return Response(data, status=http_codes.HTTP_404_NOT_FOUND)
> ```
> 
> The main benefit is it looks like exception handling, and uses the exact same exception matching as typical (as opposed to similar if branching on isinstance()). And log.exception() and/or logging calls with exc_info/stack_info works as expected [that may be okay in the isinstance() case as well, haven't tested...]

It's not expected for `exception_handler` to raise an exception it received as parameter. Moreover this approach is confusing. NEVER do such things in production code.

galaxy/api/base.py
alikins: That makes sense. I see the `detail` normalizing in the various APIException subclass __init__ now. So while the __init__ might get a more structured `detail` passed in, the instances `.detail` attribute are always normalized (the exception_handler never sees the original `detail` arg passed to the __init__s).

galaxy/api/base.py
alikins: The extracted method could be unit tested directly, but if we need to extract it later we can.

PR #1761, opened 2019-04-23
Use absolute urls in api/v2 endpoints

### LINKED ISSUE #1757: Update api/v2 urls from relative to absolute
This includes next, previous, and any urls built from `reverse()` in serializers and custom fields.

awcrosby:
Closes #1757

Updates collection, collection version, collection import, and download urls from relative to absolute.

### DIFF
--- a/galaxy/api/fields.py
+++ b/galaxy/api/fields.py
@@ -18,6 +18,7 @@
 import datetime as dt
 
 from rest_framework import serializers
+from rest_framework.reverse import reverse
 
 
 class NativeTimestampField(serializers.DateTimeField):
@@ -36,3 +37,31 @@ class NativeTimestampField(serializers.DateTimeField):
             return None
         value = super().to_internal_value(value)
         return value.astimezone(dt.timezone.utc).timestamp()
+
+
+class NamespaceObjectField(serializers.Field):
+    """Return namespace object for a serializer field."""
+    def to_representation(self, value):
+        return {
+            'id': value.pk,
+            'href': reverse(
+                'api:namespace_detail',
+                kwargs={'pk': value.pk},
+                request=self.parent.context.get('request'),
+            ),
+            'name': value.name,
+        }
+
+
+class VersionUrlField(serializers.Field):
+    """Return version detail url under collection namespace and name."""
+    def to_representation(self, value):
+        return reverse(
+            'api:v2:version-detail',
+            kwargs={
+                'namespace': value.collection.namespace.name,
+                'name': value.collection.name,
+                'version': value.version,
+            },
+            request=self.parent.context.get('request'),
+        )
--- a/galaxy/api/fields.py
+++ b/galaxy/api/fields.py
@@ -47,7 +47,7 @@ class NamespaceObjectField(serializers.Field):
             'href': reverse(
                 'api:namespace_detail',
                 kwargs={'pk': value.pk},
-                request=self.parent.context.get('request'),
+                request=self.context.get('request'),
             ),
             'name': value.name,
         }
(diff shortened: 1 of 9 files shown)


alikins:
@awcrosby For a request like `GET /api/v2/collections/alikins/collection_reqs_test`, could the `versions_url` include params? For example, `page_size`?

Seems like it would be useful for server to do that at some point, so was 
wondering if there will be cases where mazer and or other rest client will need to combine params from the urls in the responses with local ones. Not important at the moment, just a note noticed when the code is in my head.

alikins:
@awcrosby I did a little bit of testing this branch, and started a mazer branch updating it for full urls at https://github.com/ansible/mazer/pull/232 . It's enough for the 'install' case at the moment, but it's still a WIP.

awcrosby:
> For a request like `GET /api/v2/collections/alikins/collection_reqs_test`, could the `versions_url` include params? For example, `page_size`?

@alikins query params are supported, including `page_size` (up to a maximum). If we find use cases in the future where mazer needs filtered info, we can see if query params or a different endpoint is best.

alikins:
Since this presumably only effects v2/ api (and I guess maybe /internal at some point), wondering if there should be some indication that urls are full absolute vs the url fragment used in v1 (and in /api/). Aside from priori knowing v1 is short urls and v2 is full.

Probably not a concern at the moment, but wondering about API clients that will need to use a mix of /v1, /v2, and internal/ (I guess the webui is the primary one at this point, but thinking of things like a potential Tower integration). I don't think we have to figure it out now, if nothing else could use a custom media type or similar mechanism if it ever matters.

alikins:
@awcrosby 


> @alikins query params are supported, including `page_size` (up to a maximum). If we find use cases in the future where mazer needs filtered info, we can see if query params or a different endpoint is best.

Yup, stuff like filtering is what I was thinking. In particular query params that may be duplicated (ie, 'content_type=action;content_type=module' etc). I don't think there are any problems there, just thinking it through.

galaxy/api/fields.py
```diff
@@ -36,3 +37,31 @@ def to_internal_value(self, value):
...
+
+class NamespaceObjectField(serializers.Field):
+    """Return namespace object for a serializer field."""
+    def to_representation(self, value):
+        return {
+            'id': value.pk,
+            'href': reverse(
+                'api:namespace_detail',
+                kwargs={'pk': value.pk},
+                request=self.parent.context.get('request'),
```
cutwater: Why `self.parent.context` here? Fields has `self.context` attribute.

galaxy/api/fields.py
```diff
@@ -36,3 +37,31 @@ def to_internal_value(self, value):
...
+    """Return version detail url under collection namespace and name."""
+    def to_representation(self, value):
+        return reverse(
+            'api:v2:version-detail',
+            kwargs={
+                'namespace': value.collection.namespace.name,
+                'name': value.collection.name,
+                'version': value.version,
+            },
+            request=self.parent.context.get('request'),
```
cutwater: Ditto

galaxy/api/v2/views/collection_version.py
```diff
@@ -120,4 +121,5 @@ def get(self, request, pk=None, namespace=None, name=None, version=None):
                 collection__name__iexact=name,
                 version__exact=version,
             )
-        return redirect(version.get_download_url())
+        absolute_path = request.build_absolute_uri(version.get_download_url())
```
cutwater: We don't need to build absolute URLs for redirects. Only for API json values.

PR #1762, opened 2019-04-23
Add collections to namespace detail page

newswangerd:
image

### DIFF
--- a/galaxyui/src/app/authors/detail/author-detail.component.html
+++ b/galaxyui/src/app/authors/detail/author-detail.component.html
@@ -199,7 +199,7 @@
     <div class="row repository-pagination" *ngIf="!authorNotFound">
         <div class="col-sm-12">
             <div class="pagination"
-                *ngIf="repositories && repositories.length && paginationConfig.totalItems > paginationConfig.pageSize">
+                *ngIf="paginationConfig.totalItems > paginationConfig.pageSize">
                 <pfng-pagination
                     [config]='paginationConfig'
                     (onPageSizeChange)="handlePageSizeChange($event)"
--- a/galaxyui/src/app/authors/detail/author-detail.component.html
+++ b/galaxyui/src/app/authors/detail/author-detail.component.html
@@ -112,19 +112,19 @@
             </div>
             <div class="metric-box mid-font pull-left list-content">
                 <div>
-                    <b>{{item.latest_version.content_counts.module}}</b>
+                    <b>{{item.latest_version.content_summary.type_count.module}}</b>
                     Modules
                 </div>
                 <div>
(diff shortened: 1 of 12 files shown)


galaxyui/src/app/authors/detail/author-detail.component.html
```diff
@@ -172,7 +199,7 @@
     <div class="row repository-pagination" *ngIf="!authorNotFound">
         <div class="col-sm-12">
             <div class="pagination"
-                *ngIf="items && items.length && paginationConfig.totalItems > paginationConfig.pageSize">
+                *ngIf="repositories && repositories.length && paginationConfig.totalItems > paginationConfig.pageSize">
```
ironfroggy: This looks like we'll only include pagination if there are repositories, but we want to show pagination if there are repositories AND/OR collections, right?

galaxyui/src/app/authors/detail/author-detail.component.ts
```diff
@@ -356,7 +369,7 @@ export class AuthorDetailComponent implements OnInit {
     }
 
     private prepareRepositories() {
-        this.items.forEach((item: Repository) => {
+        this.repositories.forEach((item: Repository) => {
```
ironfroggy: Is there no similar need for a `prepareCollections()` helper?

galaxyui/src/app/utilities/score/score.component.ts
```diff
@@ -1,9 +1,12 @@
 import { Component, Input, OnInit } from '@angular/core';
 
-class RepoScore {
+interface IRepoScore {
     community_score: any;
-    quality_score: any;
+    quality_score?: any;
```
ironfroggy: Why is the quality score now optional?

newswangerd: This component was originally supposed to accept a `Repository` type which has `community_score`, `quality_score` and `community_survey_count` attributes. Collections also need to be able to use this component now and collections don't a `quality_score` attribute.

Modifying it so that it either accepts a `quality_score` or `latest_version` object was the easiest way to make it so it accepts collections and repositories.

galaxyui/src/app/authors/detail/author-detail.component.ts
newswangerd: I really hate this pattern because the `prepareRepositories()` function adds new attribute to the `Repository` type. The result is that you end up with a list of repositories that don't actually match the correct type definition and that makes it a lot hard to debug and extend the UI. If I had more time I'd go back and fix the legacy code on this page so that it doesn't mess with the `Repository` type.

We also don't really need a `prepareCollections()` function because the backend already gives us all the data we need for displaying collections.

PR #1768, opened 2019-04-26
UI for collections on search page

newswangerd:
Rebased on #1807. Mostly finished, just waiting on some minor changes to the search API params.

image


ironfroggy:
Here's some cosmetic issues I noticed:

- The user column on the left of the result rows are a different size for collections and repositories
- The titles aren't the same font/size/weight
- The description for collections has a label, the repositories don't. Kind of seems redundant on the collections?

I don't know if any of those are _blockers_, but if they aren't resolved for this PR so the work can get merged and shared, I'll capture it in a ticket. Combining the two types in the same result set shouldn't be so jarring for the user.

newswangerd:
@ironfroggy I'll fix point 2 and 3 as part of this PR. I think point 1 would be better served in another PR. The reason being that the list of repos are actually in a patternfly list (which comes with all of patternfly's styling BS) and so to fix this I need to rip out the patternfly list and replace it with something else. This could be kind of time consuming and has the potential to introduce even more CSS bugs and I don't want those changes to keep this from getting merged.

ironfroggy:
@newswangerd 👍 I'll look for an update with changes for those two, and at that point will file a separate ticket to track the other issue more minor issue

Issue #1740, opened 2019-04-14
Platforms card is too narrow with reduced-width window


westurner:
##### SUMMARY
The platforms card is too narrow such that I have to horizontal scroll to see the overflow when I have a browser sidebar open on my tiny laptop.

##### STEPS TO REPRODUCE

1. https://galaxy.ansible.com/geerlingguy/awx
2. Open a browser sidebar; such as Firefox Tree Style Tabs. (Screenshots attached)
3. Read the platforms card

##### EXPECTED RESULTS

I could read the platform versions without horizontal scrolling.

##### ACTUAL RESULTS
``col-sm-3``:
![Screenshot from 2019-04-14 10-49-41](https://user-images.githubusercontent.com/50891/56094769-20d52d80-5ea4-11e9-9526-6c794ca15ab2.png)

Edited to ``col-sm-4``:
![Screenshot from 2019-04-14 10-49-23](https://user-images.githubusercontent.com/50891/56094771-2599e180-5ea4-11e9-85a9-0c7e0aeea227.png)

Oddly, https://github.com/ansible/galaxy/blob/devel/galaxyui/src/app/content-detail/content/roles/roles.component.html already has ``col-sm-4``; but that's not what I see on the live site; so I must have the wrong template?

westurner:
Btw, Here's a 180px sidebar instead of the current 250px
Current:
![Screenshot from 2019-04-14 11-10-00](https://user-images.githubusercontent.com/50891/56094964-e10f4580-5ea5-11e9-87f6-9da3d6d021f2.png)
180px:
![Screenshot from 2019-04-14 11-07-41](https://user-images.githubusercontent.com/50891/56094936-bb823c00-5ea5-11e9-9278-2d4b08ef4030.png)
![Screenshot from 2019-04-14 11-08-17](https://user-images.githubusercontent.com/50891/56094937-bb823c00-5ea5-11e9-8511-758619ab00ff.png)

190 or 200 might look better with the header nav logo.

westurner:
Thanks!

PR #1779, opened 2019-05-02
Collapse navigation on smaller screens

newswangerd:
Collapses the navigation down to just display the icons on screens that are less than 1300 pixels. Wide.

Unfortunately there doesn't seem to be a way to get tooltips to work on the nav bar, so we don't have a way to display the name of the page when the nav bar is collapsed other than re-expanding it.

Screen Shot 2019-05-02 at 4 44 09 PM

### DIFF
--- a/galaxyui/src/app/app.component.html
+++ b/galaxyui/src/app/app.component.html
@@ -1,6 +1,11 @@
-<div id="verticalNavLayout" class="layout-pf layout-pf-fixed faux-layout" style="background-color: white;"
-    (window:resize)="onResize($event)">
+<div
+    id="verticalNavLayout"
+    class="layout-pf layout-pf-fixed faux-layout"
+    style="background-color: white;"
+    (window:resize)="onResize($event)"
+>
     <pfng-vertical-navigation
+        #verticalNav
         [brandAlt]="'Galaxy'"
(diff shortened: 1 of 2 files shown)


newswangerd:
Partial fix for #1740 as well as many other issues on small screens.

ironfroggy:
Is it not possible to achieve this by customizing the value of the breakpoints defined here:

https://github.com/patternfly/patternfly-ng/blob/84762094bdc7a478d7c6340b66e59057e78963ec/src/app/navigation/vertical-navigation/vertical-navigation.component.ts#L132-L135

Which are currently 768 for tablet and 1200 for desktop? Also, that means this PR collapsing at 1300 is collapsing _above_ what existing breakpoints treat as "desktop".

newswangerd:
@ironfroggy The tablet breakpoint causes the entire menu to disappear into the hamburger menu.

Screen Shot 2019-05-06 at 11 06 50 AM

I have no idea what the desktop breakpoint variable does for the vertical menu (it doesn't seem to be used anywhere in this component). 

I was originally going to do 1200px, but after spending some time looking at it I decided that keeping the menu expanded at that size just feels too cramped. I thought 1300 was a good size because it feel spacious enough and most displays are larger than 1300: https://www.w3schools.com/browsers/browsers_display.asp

Issue #1780, opened 2019-05-02
My Imports namespace selector broken on Chrome


ironfroggy:
##### SUMMARY
On the Chrome browser I am unable to select a non-default Namespace to view imports. This works on Firefox.

##### STEPS TO REPRODUCE
- Create an account with access to a namespace.
- Navigate to My Imports
- Try to change the Namesapce

##### EXPECTED RESULTS
I expect to see the imports for the other namespace.

##### ACTUAL RESULTS
Nothing happens. This is in the console:

```
Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
    in select (created by FormControl)
    in FormControl (created by ImportListComponent)
    in div (created by ImportListComponent)
    in div (created by ImportListComponent)
    in div (created by ImportListComponent)
    in ImportListComponent (created by MyImportsPage)
    in div (created by MyImportsPage)
    in div (created by MyImportsPage)
    in div (created by MyImportsPage)
    in MyImportsPage

Warning: React does not recognize the `initExpanded` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `initexpanded` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in div (created by ListViewGroupItem)
    in ListViewGroupItem (created by ListViewItem)
    in ListViewItem (created by ImportListComponent)
    in div (created by ListView)
    in ListView (created by ImportListComponent)
    in div (created by ImportListComponent)
    in div (created by ImportListComponent)
    in ImportListComponent (created by MyImportsPage)
    in div (created by MyImportsPage)
    in div (created by MyImportsPage)
    in div (created by MyImportsPage)
    in MyImportsPage
```

newswangerd:
Just a small note: the `initExpanded` error is from patternfly-react: https://github.com/patternfly/patternfly-react/issues/1746

PR #1782, opened 2019-05-03
Fix namespace selector in Chrome

newswangerd:
Fixes: #1780


galaxyui/src/app/react/containers/my-imports.tsx
```diff
@@ -342,7 +342,12 @@ export class MyImportsPage extends React.Component<IProps, IState> {
         );
     }
 
-    private selectedNamespace(ns: Namespace) {
+    private selectedNamespace(nsID: number) {
+        // For some reason the value passed back by the react component isn't
```
ironfroggy: Because all attributes become strings on the DOM

PR #1785, opened 2019-05-03
Set default value for repo import status on my content.

### LINKED ISSUE #1776: My Content not always displaying content
On the My Content page for my user on galaxy-dev, when I expand my namespace to see the repositories and collections, the content flashes briefly and then does not display. This does not occur on my local testing environment, nor does it occur for other namespaces (including those with both repositories and collections, alikins for example)

https://galaxy-dev.ansible.com/my-content/namespaces

image

Browser console shows:
```
Invariant Violation: "Minified React error #143; visit https://reactjs.org/docs/error-decoder.html?invariant=143 for the full message or use the non-minified dev environment for full errors and additional helpful warnings. "
```

newswangerd:
I think this will probably fix #1776 but I can't replicate the issue, so I need @awcrosby to test it for me.

### DIFF
--- a/galaxyui/src/app/react/components/my-content/repo-list-item.tsx
+++ b/galaxyui/src/app/react/components/my-content/repo-list-item.tsx
@@ -95,7 +95,7 @@ export class RepoListItem extends React.Component<IProps, {}> {
             this.props.namespace.id
         }?type=repository&name=${repo.name}`;
 
-        let importElement;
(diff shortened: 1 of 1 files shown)


awcrosby:
@newswangerd since the #1776 issue only occurs on galaxy-dev I cannot test in my local environment. If you feel comfortable merging, I can redeploy galaxy-dev and test there

PR #1786, opened 2019-05-03
Add 'test/quick' and 'dev/test/quick' targets

alikins:
The 'quick' tests mostly just also add the
pytest option '--reused-db'. They also turn off
test coverage.

And use PYTEST_TARGET and PYTEST_ARGS with pytest.

PYTEST_TARGET is used in 'pytest $(PYTEST_TARGET) and
defaults 'galaxy', but could be
'galaxy/main/tests/test_whatever.py::test_stuff'

PYTEST_ARGS is empty, but can be used for any pytest
args ('-vvv' or '--log-level=DEBUG' etc)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,9 @@ GALAXY_RELEASE_TAG ?= latest
 GALAXY_VENV=/usr/share/galaxy/venv
 DOCKER_COMPOSE=docker-compose -p galaxy -f ./scripts/docker/dev/compose.yml
 
+PYTEST_TARGET ?= galaxy
+PYTEST_ARGS ?=
+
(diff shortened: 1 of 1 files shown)


ironfroggy:
I asked in chat if we could just add `--reuse-db` as the default instead of adding a new target. If the resused DB ever breaks, it can be wiped and rebuilt, so as a default it feels safe and a time saver. What do you think @alikins ?

ironfroggy:
@alikins Are we worried about race conditions with this increment, and if so, do you want to try one of the atomic increment methods detailed here: https://stackoverflow.com/a/1599090

Issue #1769, opened 2019-04-29
Download count not getting incremented for collections


newswangerd:
Running `mazer install` should increment the download count for collections.

ironfroggy:
The way this was implemented for repositories was pretty inaccurate, it increment every time a repo was just viewed. Hopefully we can do it accurately this time?

PR #1789, opened 2019-05-03
Rev Collection download_count on /artifact/ get

alikins:
Rev Collection download_count on /artifact/ GET

CollectionArtifactView.get() increments the artifacts
related Collection's download_count.

Add a unit test to verify.

Work in progress at the moment

- [x] Add Collection download count for 'mazer' download_url hits
  - [x] inc count on download_url hits
  - [x] exclude any non-mazer hits, including from the wui
- [x] ~Determine if the count should be against a Collection entirely, or a CollectionVersion, or both~
- [x] cleanup debug logging
- [x] more tests (for mazer vs wui stuff when it exists)

Aside from the unit test and some manual testing with mazer, this hasn't been
tested much (ie, haven't tested with concurrent hits to the same /artifact/ url etc).

### DIFF
--- a/galaxy/api/v2/tests/test_collection_version_views.py
+++ b/galaxy/api/v2/tests/test_collection_version_views.py
@@ -14,7 +14,6 @@
 #
 # You should have received a copy of the Apache License
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
-import logging
 
 from django.contrib.auth import get_user_model
 from pulpcore.app import models as pulp_models
@@ -24,8 +23,6 @@ from rest_framework.test import APITestCase
 from galaxy.main import models
 
 UserModel = get_user_model()
-
-log = logging.getLogger(__name__)
 
 mazer_user_agent = \
     'Mazer/0.4.0 (linux; python:3.6.8) ansible_galaxy/0.4.0'
@@ -68,7 +65,6 @@ class TestCollectionArtifactView(APITestCase):
 
     def test_download_count(self):
         download_count_before = self.collection.download_count
-        log.debug('download_count_before: %s', download_count_before)
 
         url = '/api/v2/collections/mynamespace/mycollection' \
             + '/versions/1.2.3/artifact/'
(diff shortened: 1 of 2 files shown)


ironfroggy:
wrt Collection vs CollectionVersion, we should count per-version so we can display individual or cumulative.

chouseknecht:
Fixes #1769

alikins:
Note, this needs mazer of rev 868f05ebcb16ea84a451e4b137814ee01bcaab41 or later. Otherwise the artifact download request doesn't have the same user-agent as the other requests.

galaxy/api/v2/tests/test_collection_version_views.py
```diff
@@ -44,6 +50,54 @@ def setUp(self):
...
+        response = self.client.get(url,
+                                   HTTP_USER_AGENT=mazer_user_agent)
+
+        assert response.status_code == http_codes.HTTP_302_FOUND
+
+        self.collection.refresh_from_db()
+
+        assert self.collection.download_count == download_count_before + 1
+
+    def test_download_count_non_mazer_user_agent(self):
```
ironfroggy: The extra logging isn't really adding anything, because pytest inspects the frame when an assert fails and reports the values of the variables in the assert expression. So if the assert on the `download_count` vs `download_count_before` fail, it'll report both values.

The logging doesn't hurt, obviously, just an FYI

alikins:
> @alikins Are we worried about race conditions with this increment,

I am kind of worried about that. The changes here were based on how download count was handled by roles on the theory that is workable. But I'm not actually sure it won't become a problem. I've seen this kind of counter/increment on ORM objects causes problems before (though, that was in a system with very long open transactions and mysql with 'optimisitic row locking' iirc, so may not apply here)

I can try some of the ideas suggested at https://stackoverflow.com/a/1599090.  Any suggestions on how to test the concurrent use of this?

alikins:
I suspect the download_count on the Collection (Repository, etc) isn't the right approach, and it may end up being better to track that external to that object/table.

In the meantime, looking into the F() and other incrementing methods.

Long term, if we assume that info does need to live in the db (as opposed to metrics aggregates or even logging), could just move it to some sort of download table (or a more generic metrics table).

Maybe something like:
 https://github.com/pennersr/django-trackstats
https://github.com/rmaceissoft/django-user-metrics
https://github.com/frankwiles/django-app-metrics

Seems like you could also potentially do it with a middleware module, though not sure middleware apps are expected to write to the db.

galaxy/api/v2/tests/test_collection_version_views.py
alikins: yup, true. Leftovers from test bootstrapping. I'll clean them up.

Issue #1771, opened 2019-04-30
Import page filtering by type fails to apply its default value


ironfroggy:
##### SUMMARY
The default value of the "Type" filter available on the My Imports page is "Collection", but this is not actually applied. The filter as shown does not reflect the contents shown.

##### STEPS TO REPRODUCE
- Navigate to the My Imports page
- Change the filter type from "Name" to "Type"

##### EXPECTED RESULTS
Either no default, since there is no filter by default, or for the default to be applied when it appears.

##### ACTUAL RESULTS
The Collection Type filter only applies by clicking the drop down and re-clicking the value that was already selected in the first place. Until then, no filtering is applied.

![galaxy-import-filter-default-fail](https://user-images.githubusercontent.com/34982/56982263-e8457d00-6b4e-11e9-9e3d-f715ada02005.gif)

PR #1792, opened 2019-05-06
Fix placeholder on My Imports type filter.

newswangerd:
Fixes: #1771

### DIFF
--- a/galaxyui/src/app/react/components/my-imports/import-list.tsx
+++ b/galaxyui/src/app/react/components/my-imports/import-list.tsx
@@ -42,24 +42,16 @@ interface IProps {
     setQueryParams: (filters) => void;
 }
 
-export class ImportListComponent extends React.Component<IProps, {}> {
-    appliedFilters = [] as AppliedFilter[];
-    filterConfig = {} as FilterConfig;
+interface IState {
+    selectedFilter: FilterOption;
+    filterValue: string;
+}
 
-    render() {
-        const {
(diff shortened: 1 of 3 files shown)


newswangerd:
I looked at the angular filter and realized that the placeholder text doesn't actually change when you select something from the list, so I'm going to copy that behavior here (which also fixes the issue where the filter gets set to Repository or Collection when you switch back to name)

I'm not sure what you mean when you say you can't remove filters. To go back to no filter you're supposed to use the clear all filters control or remove filters one by one. I don't think the patternfly filters are designed to be removed from the dropdown menu.

Screen Shot 2019-05-10 at 1 56 45 PM

ironfroggy:
If I select a Type, like "Collection", and then change the same dropdown back to its default, non-selected state, that should undo the Type=Collection filtering, too. I can click "Clear All Filters" but this is still a misbehaving dropdown if I can change its value back to return to the previous state.

newswangerd:
Well, this is how they're implemented in patternfly ¯\_(ツ)_/¯

http://patternfly-react.surge.sh/patternfly-3/index.html?selectedKind=patternfly-react%2FForms%20and%20Controls%2FFilter&selectedStory=Filter&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

https://www.patternfly.org/patternfly-ng/#/filters

PR #1795, opened 2019-05-06
Fix scroll to end of log bug on my imports

### LINKED ISSUE #1770: "Scroll to end of log" breaks import navigation on My Imports page
##### SUMMARY
Using the "Scroll to end of log" icon to quickly go to the end of an import log causes navigation to any other imports on the My Import page to break, wiping the contents of the page and leaving the page in a stuck state. Clicking to go back to "My Imports" does not work, only refreshing or navigating away and back restores functionality.

##### STEPS TO REPRODUCE
- Import more than one repository or collection
- Scroll to the bottom of one import log with the down arrow icon at the top of the log
- Try to navigate to the other import log by clicking it in the list to the left of the log

##### EXPECTED RESULTS
I expect to go read the other import log available to me

##### ACTUAL RESULTS
The contents of the page are gone. Only the top bar and left-hand navigation remain.

galaxy-import-nav-failure

newswangerd:
Fixes #1770

### DIFF
--- a/galaxyui/src/app/react/components/my-imports/import-console.tsx
+++ b/galaxyui/src/app/react/components/my-imports/import-console.tsx
@@ -100,13 +100,12 @@ export class ImportConsoleComponent extends React.Component<IProps, {}> {
                             </span>
                         </div>
                     ) : null}
-
-                    <div
-                        className='message'
-                        key={'last'}
-                        ref={this.lastImport}
-                    />
                 </div>
+                <div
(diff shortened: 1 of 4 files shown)


newswangerd:
Changes:
- When a new import is selected, page always scrolls to top
- While an import is running the button toggles following logs
- When an import is finished running follow logs is automatically toggled off
- When an import is not running the button scrolls to the end of the logs

Issue #1783, opened 2019-05-03
Allow collections to be added to namespaces with no provider namespace.


newswangerd:
![Screen Shot 2019-05-03 at 10 32 00 AM](https://user-images.githubusercontent.com/6063371/57144292-c20e2000-6d8e-11e9-84b5-d6d1b83d9e8e.png)

Add content button is greyed out when no provider namespace is configured for a namespace.

PR #1796, opened 2019-05-06
Allow collections to be uploaded on namespaces with no providers.

newswangerd:
fixes: #1783


newswangerd:
> There is another bug that is related to this: you can't actually view any of the content for a non-provider namespace, so even if you can publish something you can't actually navigate to it.

I think I'm going to fix this in a separate PR. I'm planning on reworking this page to use the same combined API endpoint that we use the for namespace detail page. The fix for this will get included as part of that PR.

ironfroggy:
Since this was a pre-existing related but separate bug, I'll agree.

Issue #1750, opened 2019-04-16
The My Content page's collection list almost lays cuts off collection names


ironfroggy:
##### SUMMARY
The My Content page's collection list almost lays cuts off collection names.

##### STEPS TO REPRODUCE
Upload a few collections and look at your list of collections at /my-content/namespaces

##### EXPECTED RESULTS
I expected to be able to tell the difference between each collection in my list

##### ACTUAL RESULTS
I can't tell which is which, because very little space is given to the name that becomes cut off.

![collecttableshortcolumn](https://user-images.githubusercontent.com/34982/56244717-950dfd80-606b-11e9-9232-4099becae1c4.png)

@newswangerd and I talked about this and discussed this layout which requires minimal adjustment:

![collectiontable](https://user-images.githubusercontent.com/34982/56244732-a6570a00-606b-11e9-8df7-860ed19d0683.png)

PR #1797, opened 2019-05-06
Rejigger my content to do a better job of showing descriptions.

newswangerd:
Fixes:  #1750

Screen Shot 2019-05-06 at 3 29 08 PM


ironfroggy:
content-strange-alignment

I think this alignment is a little strange. Would it be possible to align these elements with the content name at the top of each row?

newswangerd:
> content-strange-alignment
> 
> I think this alignment is a little strange. Would it be possible to align these elements (icon, button, and score) with the content name at the top of each row?

I can look into it, but I really hate having to override the patternfly CSS because it's likely that it will make debugging the CSS much harder in the future.

I think I may just build my own list for this since we aren't really using a lot of the patternfly list features here anyway.

PR #1820, opened 2019-05-15
Refactor my content to use the combined repo/collection API

newswangerd:
My content now displays collections and repositories.

Screen Shot 2019-05-15 at 9 19 48 AM

With an option to filter out one or the other.
Screen Shot 2019-05-15 at 9 21 02 AM


ironfroggy:
Still shows the Repositories count at the bottom even if there are none on the page. it should only shown on pages where the count is >0

PR #1824, opened 2019-05-16
Enforce unique names across collections, and repository roles

awcrosby:
- On collection import, check for name conflict with repository roles
  - the check occurs in the API collection list view post.
- On repository import, check for name conflict with collections
  - the check occurs in the worker, which is called from multiple API endpoints (repository, notifications, import task). The check occurs after the importer has processed the repository and potentially renamed the repository based on the `meta/main.yml` `role_name` attribute

### DIFF
--- a/galaxy/api/v2/views/collection.py
+++ b/galaxy/api/v2/views/collection.py
@@ -153,7 +153,7 @@ class CollectionListView(base.APIView):
     def _check_role_name_conflict(self, ns, name):
         roles = models.Content.objects.filter(
             content_type__name=constants.ContentType.ROLE,
-            namespace=ns,
+            repository__provider_namespace__namespace=ns,
             name=name,
         )
         if not roles:
--- a/galaxy/api/v2/views/collection.py
+++ b/galaxy/api/v2/views/collection.py
@@ -29,6 +29,7 @@ from rest_framework import status as http_codes
 from pulpcore.app.serializers import ArtifactSerializer
 from pulpcore.app import models as pulp_models
 
+from galaxy import constants
 from galaxy.api import base
 from galaxy.api import exceptions
 from galaxy.api.v2 import serializers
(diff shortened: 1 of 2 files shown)


galaxy/api/v2/views/collection.py
```diff
@@ -143,6 +150,21 @@ def _check_namespace_access(self, namespace, user):
                 'the namespaces you have access to.'
             )
 
+    def _check_role_name_conflict(self, ns, name):
+        roles = models.Content.objects.filter(
+            content_type__name=constants.ContentType.ROLE,
+            namespace=ns,
```
awcrosby: Options to consider here:
- `namespace=ns`
- `repository__provider_namespace__namespace=ns`

ironfroggy:
galaxy-duplicate

I was still able to create a duplicate by publishing a collection named "just_a_role" and then importing my `orionuser1.just-a-role` repository.

In the other direction, it worked fine. I wasn't able to publish the collection if I imported the repository first.

galaxy/api/v2/views/collection.py
chouseknecht: We have `namespace` on the Content object for convenience, but I think to be accurate, we should use `provider_namespace__namespace`.

chouseknecht:
Following up on comment from @ironfroggy, this looks like a case of converting `-` to `_`. We need to account for that in the duplicate check.

awcrosby:
@ironfroggy can you confirm this occurs after you switch to this branch and restart the celery workers? This scenario should be taken care of

> I was still able to create a duplicate by publishing a collection named "just_a_role" and then importing my orionuser1.just-a-role repository.

galaxy/api/v2/views/collection.py
awcrosby: update was made to `repository__provider_namespace__namespace`

Issue #1607, opened 2019-03-01
Clean up API requests on content details


newswangerd:
The content details page currently has to make multiple paged API requests to load versions, content and. We need to redesign the API and page so that it makes fewer, more condensed API calls.

Issue #1810, opened 2019-05-09
Inconsistencies in handling of namespace casing between legacy and collection versions


ironfroggy:
##### SUMMARY
Handling of namespace casing for collections is inconsistent with current treatment and actually broken for some namespaces. Specifically these are non-username namespaces, since usernames from Github logins are always lowercase in the first place.

Here's a summary of the issues found:
- Mazer refuses to build namespaces with uppercase characters, although these are valid namespaces and many such namespaces are in active use today for repositories.
- Collections built with lowercase versions of mixed-case namespaces cannot be uploaded by the UI, because the UI fails to locate the namespace by that name.
- It isn't clear yet if collections eventually built and published to mixed-case namespaces (when this is fixed) will be installed properly by Mazer. I can't test this until I can publish them in the first place.
- If a namespace is renamed just to fix the casing, which is a request we've had, all existing collections are no longer accessible, even though the URLs to access the collection detail page are themselves case insensitive. For example, if I publish a collection to `http://127.0.0.1:8000/collections/orionuser1ns/collection_simple_role_zmaiwwfk` I can also bring up that same content at `http://127.0.0.1:8000/collections/OrionUser1NS/collection_simple_role_zmaiwwfk`. If I then I rename the collection from `"orionuser1ns"` to `"OrionUser1NS"`, the same URL no is a 404.

This is a collection is issue related to the namespace casing and I'm filing it partly to document any information that can be provided or decisions around this. I don't know what behaviors we see in namespaces today are deliberate versus coincidental, so maybe we don't _want_ to support mixed-case namespaces!

But the fact remains that we already have them, that people are already using them, and that we already have established patterns of behavior for them. So, what's the right way to handle this with regard to collections?

chouseknecht:
Related to #1805

alikins:
@ironfroggy 


> * It isn't clear yet if collections eventually built and published to mixed-case namespaces (when this is fixed) will be installed properly by Mazer. I can't test this until I can publish them in the first place.

My suspicion is that mazer would find it, download it, extract/install it, then raise errors when it 'reloads' the installed collection and applying the CollectionInfo/galaxy.yml validation.  

I don't think it would fail until after it extracted the artifact to the collections_path, since the artifact extraction code doesn't validate the content. ie, the namespace lowercase restriction wouldn't be validated until it needs to create a mazer CollectionInfo object, which it avoids doing during the extract phase. 

[...]

PR #1825, opened 2019-05-16
Routes collections and repositories via /namespace/content-name

newswangerd:
Fixes: #1607


galaxy/api/internal/views/combined.py
```diff
@@ -115,3 +116,58 @@ def _validate_order(self, order):
...
+class CombinedDetail(base.APIView):
+    '''
+    This is intendended to provide all of the information for the content
+    detail pages. For repos, it returns the repository, namespace and list of
+    content items.
+
+    For collections it returns a collection object
+    '''
+    def get(self, request):
+        namespace = request.GET.get('namespace', None)
```
chouseknecht: Question. How are we handling case sensitivity. Specifically, thinking about #1810.

chouseknecht: Just reviewed #1824. Do we potentially have the same issue here, where we need to convert `-` to `_` before performing the lookup?

ironfroggy: Case is insensitive, I think it could make sense to incorporate the `-` 👉`_` bit, if you want to basically let github URL paths work here for consistency, but does that make sense for collections where they don't have to match repository names at all?

newswangerd: I think making the query case insensitive is fine.

I don't think we should perform a `-` -> `_` conversion. We still have a lot of namespaces with `-` in the name and a change like this would make all of those pages impossible to find.

PR #1840, opened 2019-05-21
Validation of content name format

### LINKED ISSUE #1836: Collection import doesn't prevent invalid role names ('test-role-a')
## Bug Report 
For ansible 2.8, the collection role loader requires that roles in a collection meet the same naming restrictions as collection namespace, names, or plugins. ie, no dashes like in 'test-role-a'.

The lint steps at collection import should flag a role with invalid name as invalid. (And likely, 'mazer build' should do the same)

Attempting to use a role with '-' in the name from an ansible playbook

```
ERROR! the role 'alikins.ansible_testing_content.test-role-a' was not found in /home/adrian/src/testcollections/ansible-testing-content/tests/roles:/home/adrian/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:/home/adrian/src/testcollections/ansible-testing-content/tests

The error appears to be in '/home/adrian/src/testcollections/ansible-testing-content/tests/test.yml': line 6, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  roles:
    - alikins.ansible_testing_content.test-role-a
      ^ here
```

Collection is installed:

```
(mazer040test) [newswoop:F29:ansible-testing-content (master % u=)]$ ansible-config dump | grep COLLECTIONS
COLLECTIONS_PATHS(default) = ['/home/adrian/.ansible/collections', '/usr/share/ansible/collections']
(mazer040test) [newswoop:F29:ansible-testing-content (master % u=)]$ ls ~/.ansible/collections/ansible_collections/testing/ansible_testing_content/roles/test-role-a
defaults  handlers  meta  tasks  tests  vars
```

If I copy that role to a name without dashes 'testrolea', then it works:
```
(mazer040test) [newswoop:F29:ansible-testing-content (master % u=)]$ cat tests/test.yml
---
- name: Example playbook
  hosts: localhost
  gather_facts: false
  roles:
    - alikins.ansible_testing_content.testrolea
(mazer040test) [newswoop:F29:ansible-testing-content (master % u=)]$ ansible-playbook -i tests/inventory  tests/test.yml 

PLAY [Example playbook] 
***********************************************************************

PLAY RECAP 
***********************************************************************
```

##### STEPS TO REPRODUCE

- `mazer install testing.ansible_testing_content`
- create a playbook that tries to use test-role-a

``` yaml
---
- name: Example playbook
  hosts: localhost
  gather_facts: false
  roles:
    - testing.ansible_testing_content.test-role-a
```
- run the playbook `ansible-playbook test.yml`

##### EXPECTED RESULTS

`mazer build` and/or collection import to fail

awcrosby:
Fixes #1836

Add validation for role and plugin content names for collections.

~~Replace regex used for repository roles to be the same used for collection roles.~~ Repo role name validation will be kept as-is to allow legacy roles to continue to be imported.

Import fails with invalid content names.

### DIFF
--- a/galaxy/importer/loaders/base.py
+++ b/galaxy/importer/loaders/base.py
@@ -18,7 +18,10 @@
 import abc
 import logging
 import os
+import re
 
+from galaxy import constants
+from galaxy.importer import exceptions as exc
 from galaxy.importer.utils import readme as readmeutils
 from galaxy.importer.utils import lint as lintutils
 from galaxy.worker import logutils
@@ -46,10 +49,25 @@ class BaseLoader(metaclass=abc.ABCMeta):
 
         self.log = logutils.ContentTypeAdapter(
             logger or default_logger, self.content_type, self.name)
+        self._validate_name()
 
     @property
     def path(self):
         return os.path.join(self.root, self.rel_path)
+
+    def _validate_name(self):
+        """Validate content name if one found from directory path.
+
+        A repository role exists at root, so content name will not exist yet.
+        """
+        if not self.name:
+            return
+        if not re.match(constants.NAME_REGEXP, self.name):
+            raise exc.ContentNameError(
+                f"{self.content_type} has invalid name: '{self.name}', "
+                "expecting name to contain only lowercase "
+                "alphanumeric characters or '_'"
+            )
 
     def make_name(self):
         """Returns content name if it can be generated from it's path.
--- a/galaxy/worker/importers/base.py
+++ b/galaxy/worker/importers/base.py
@@ -18,7 +18,6 @@
 import logging
 import re
 
-from galaxy import constants
 from django.utils import timezone
 from django.core.exceptions import ObjectDoesNotExist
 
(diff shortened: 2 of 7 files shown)


galaxy/worker/importers/base.py
```diff
@@ -63,10 +64,12 @@ def make_content(self):
             original_name = self.data.original_name
 
         # Check name
-        if not re.match(r'^[a-zA-Z0-9_-]+$', name):
+        if not re.match(constants.NAME_REGEXP, name):
```
alikins: This is for traditional style roles import via github Repository()s right?

I don't think I'd enforce it for trad roles (yet?). Maybe a message or note about 
how the role name may not be compatible with collections, but let it import?

alikins: Or is this used by both trad roles and roles-in-a-collection?

galaxy/importer/loaders/base.py
```diff
@@ -46,11 +49,26 @@ def __init__(self, content_type, path, root, logger=None):
...
         return os.path.join(self.root, self.rel_path)
 
+    def _validate_name(self):
+        """Validate content name if one found from directory path.
+
+        A repository role exists at root, so content name will not exist yet.
+        """
+        if not self.name:
+            return
+        if not re.match(constants.NAME_REGEXP, self.name):
```
alikins: Not something to address in this pr, but I've found that to raise more specific error messages, I had to split this regex up into smaller parts. ie, for things like the leading numbers, double underscores, etc.

galaxy/worker/importers/base.py
awcrosby: This module is only executed for a traditional repository role. This change to the repository role regex was removed from this PR.

galaxy/importer/loaders/base.py
awcrosby: Having the error message point users to the right place in documentation regarding the formatting details might be a consideration rather than multiple specific error messages.

awcrosby:
Closing and reopening in attempt to trigger CI

PR #1841, opened 2019-05-22
Enable collection deprecation

newswangerd:
Fixes: #1643

Todo:
- Need to add deprecated flag in search and namespace detail. This will be done as part of the search UI PR to make things easier to rebase.

My Content
Screen Shot 2019-05-22 at 9 10 32 AM

Content Detail
Screen Shot 2019-05-22 at 9 10 46 AM

### DIFF
--- a/galaxy/api/internal/serializers/collections.py
+++ b/galaxy/api/internal/serializers/collections.py
@@ -112,3 +112,10 @@ class CollectionDetailSerializer(serializers.ModelSerializer):
             hidden=False
         )
         return VersionListSerializer(versions, many=True).data
+
+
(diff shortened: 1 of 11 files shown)


galaxy/api/internal/serializers/collections.py
```diff
@@ -112,3 +112,10 @@ def get_all_versions(self, obj):
             hidden=False
         )
         return VersionListSerializer(versions, many=True).data
+
+
+class CollectionUpdateSerializer(serializers.ModelSerializer):
```
cutwater: Optional: From REST API design standpoint for partial updates `PATH` is preferred over `PUT`. `PUT` method is usually used for sending full object state.

PR #1850, opened 2019-05-28
Implement artifact download and info views

cutwater:
* Artifact download view performs redirect to external URL if
  external S3 storage is used and serves files directly in DEBUG mode.
  Other configurations are not supported and cause error.
  An appropriate `ResponseContentDisposition` parameter is addeto to S3 URL.
  Endpoint: `/download/<filename>`
* Artifact info view returns artifact metadata (i.e.`filename`, `size`,
  `sha256`, `download_url`.
  NOTE: `filename` represents public filename and does not reflect
  actual storage path.
  Endpoints:
    - `/api/v2/collections/<namespace>/<name>/versions/<version>/artifact/`
    - `/api/v2/collection-versions/<id>/artifact/`

Fixes: #1847

### DIFF
--- a/galaxy/api/download/tests/test_download_views.py
+++ b/galaxy/api/download/tests/test_download_views.py
@@ -1,0 +1,128 @@
+# (c) 2012-2019, Ansible by Red Hat
+#
+# This file is part of Ansible Galaxy
+#
+# Ansible Galaxy is free software: you can redistribute it and/or modify
+# it under the terms of the Apache License as published by
+# the Apache Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# Ansible Galaxy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# Apache License for more details.
+#
+# You should have received a copy of the Apache License
+# along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+import importlib
+from urllib import parse as urlparse
+from unittest import mock
+
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.contrib.auth import get_user_model
+from django.test import override_settings
+from pulpcore.app import models as pulp_models
+from rest_framework import status as http_codes
+from rest_framework.test import APITestCase
+
+from galaxy.main import models
+
+
+UserModel = get_user_model()
+
+
+class TestArtifactDownloadView(APITestCase):
+    download_url = '/download/{filename}'
+    mazer_user_agent = 'Mazer/0.4.0 (linux; python:3.6.8) ansible_galaxy/0.4.0'
+
+    @classmethod
+    def setUpClass(cls):
+        # Parameters of `storages.backends.s3boto3.S3Boto3Storage` class are
+        # initialized as class-attributes on module load.
+        # Because `storages.backends.s3boto3` module can be imported even
+        # before S3 specific test is executed overriding parameters locally
+        # is not effective.
+        # Thus we import `storages.backends.s3boto3` module before any
+        # request to download view is called.
+        super().setUpClass()
+        settings = dict(
+            AWS_ACCESS_KEY_ID='key',
+            AWS_SECRET_ACCESS_KEY='secret',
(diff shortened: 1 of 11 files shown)


galaxy/api/download/views.py
```diff
@@ -0,0 +1,80 @@
...
+        if isinstance(storage, s3boto3.S3Boto3Storage):
+            return self._s3_redirect(artifact_file, filename)
+
+        # If running in debug mode serve file directly by Django.
+        if isinstance(storage, FileSystemStorage) and settings.DEBUG:
+            return self._serve_file(artifact_file.url, filename)
+
+        raise ImproperlyConfigured(
+            'Artifacts should not be served by django application'
+            ' in production mode.')
```
ironfroggy: Is an exception that only gets raised on production a good idea?

Another failure mode could be to serve up the file so things actually work, but log an error to alert us to the problem to correct it.

galaxyui/proxy.conf.js
```diff
@@ -20,7 +20,7 @@ const PROXY_CONFIG = {
         'secure': false
     },
     "/download": {
-        'target': "http://localhost:8080",
```
ironfroggy: Was this port change intentional? it looks like it would break other dev's setups if they used `8080`

galaxy/api/download/views.py
chouseknecht: This exception makes sense to me.

cutwater: Sounds reasonable to me. Will fix.

galaxyui/proxy.conf.js
ironfroggy: Oh ok, the `localhost:8080` looked like it was a default for local dev stuff. My mistake!

alikins:
Just started testing, but seems to work with git mazer so far.

alikins:
Confirmed 'context-disposition' works with curl/httpie/chrome.

alikins:
Can/should there be a hyperlink from CollectionVersion  (GET /api/v2/collections/{namespace}/{name}/) to the artifacts url? (GET /api/v2/collections/{namespace}/{name}/versions/{version}/artifact) ?

cutwater:
@alikins I'll add it as `artifact_url`.

galaxy/api/download/tests/test_download_views.py
```diff
@@ -0,0 +1,128 @@
...
+            content=self.version, artifact=self.artifact,
+            relative_path=self.filename
+        )
+
+    @override_settings(
+        MEDIA_ROOT='',
+        DEFAULT_FILE_STORAGE='storages.backends.s3boto3.S3Boto3Storage',
+        AWS_DEFAULT_ACL=None,
+    )
+    @mock.patch('storages.backends.s3boto3.S3Boto3Storage.save')
```
ironfroggy: Where do the AWS credentials come in and is there a guarantee the tests don't run with actual S3 access in case something outside the patch tries to do something?

galaxy/api/download/views.py
```diff
@@ -0,0 +1,83 @@
...
+    def _s3_redirect(self, artifact_file, filename):
+        content_disposition = f'attachment; filename={filename}'
+        parameters = {
+            'ResponseContentDisposition': content_disposition
+        }
+        url = artifact_file.storage.url(
+            artifact_file.name, parameters=parameters)
+        return redirect(url)
+
+    def _serve_file(self, artifact_file, filename):
```
alikins: These could be defined outside the class or as static methods, but I'm not picky about that so I would leave them as is (especially since nothing external to the class uses them).

galaxy/api/download/tests/test_download_views.py
cutwater: AWS credentials are set in `setUpClass` method, due to how settings are initialized in `S3Boto3Storage` class (see my comment).

Actual S3 credentials are never stored in repository and test environment has no access to them, so it's impossible unless somebody is running tests in production environment with production config.

galaxy/api/download/views.py
cutwater: Done.

galaxy/api/download/views.py
cutwater: These functions are tightly related to the view class and not intended to be reused elsewhere, therefore these better to be kept in a view class scope.

PR #1854, opened 2019-05-29
Upgrade django rest framework to 3.9.4

cutwater:


ironfroggy:
We're planning to thumbs up/thumbs down the 3.2 launch in only a couple days... is that the right time to tick up the dependency versions and potentially introduce regression bugs we might not notice while we're busy finishing everything up? Maybe this is a good idea to do right _after_ the launch?

chouseknecht:
@cutwater 

Looks like DRF is the cause of our dependency security alert. Let's discuss during standup.

chouseknecht:
Hold until after 3.2 release.

chouseknecht:
Closing per discussion with @cutwater. This is out of date and needs revisiting.

PR #1855, opened 2019-05-29
Additional download unit tests and better errors

cutwater:
* Add negative test cases to download API endpoint.
* NotFound errors have more descriptive messages.
* Add string representation to CollectionFilename class.

### DIFF
--- a/galaxy/api/download/tests/test_download_views.py
+++ b/galaxy/api/download/tests/test_download_views.py
@@ -126,3 +126,25 @@ class TestArtifactDownloadView(APITestCase):
 
         assert response.status_code == http_codes.HTTP_200_OK
         assert response.getvalue() == b'CONTENT'
+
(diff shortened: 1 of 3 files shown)


galaxy/storage.py
```diff
@@ -0,0 +1,26 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# Apache License for more details.
+#
+# You should have received a copy of the Apache License
+# along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+from storages.backends import s3boto3
+
+
+class S3Storage(s3boto3.S3Boto3Storage):
```
ironfroggy: Is this used anywhere? Or going to be?

cutwater: This shouldn't be here. Thanks, good catch.

PR #1856, opened 2019-05-29
Update quality score details to contain error messages.

newswangerd:
image


galaxy/api/internal/tests/test_repo_or_collection_detail.py
```diff
@@ -17,11 +17,13 @@
...
 from rest_framework.test import APITestCase
 from rest_framework import status
+from pulpcore import constants as pulp_const
+from pulpcore.app import models as pulp_models
 
 from galaxy.main import models
 
 
-class RepoAndCollectionListTest(APITestCase):
+class RepoOrCollectionTest(APITestCase):
```
cutwater: Test case class names should begin with `Test`

PR #1859, opened 2019-05-29
Add search by type, fix search by deprecation

newswangerd:
- Added `type` param to search API that accepts None, collection or role
- Fixed bug where `deprecated=False` param was getting ignored.

### DIFF
--- a/galaxy/api/internal/views/search.py
+++ b/galaxy/api/internal/views/search.py
@@ -33,6 +33,12 @@ CONTENT_ONLY_FILTERS = [
     'cloud_platforms',
 ]
 
+ALLOWED_CONTENT_TYPES = [
+    'collection',
+    'role',
+    None
+]
+
 
 def _ensure_positive_int(string, field, cutoff=None):
     msg = f'{field} must be a positive integer'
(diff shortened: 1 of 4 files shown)


galaxy/api/internal/views/search.py
```diff
@@ -33,6 +33,12 @@
     'cloud_platforms',
 ]
 
+ALLOWED_CONTENT_TYPES = [
+    'collection',
+    'role',
+    None
```
cutwater: Please remove `None` here.

galaxy/api/internal/views/search.py
```diff
@@ -133,6 +145,16 @@ def get_order_by(self, request):
                 f'{repr(param)} is not a valid ordering parameter value.')
         return param, order
 
+    def get_format_type(self, request):
```
cutwater:
```
def get_format_type(self, request):
    format = request.query_params.get('type')  # Default is already None.
    if format is None:
        return None
    format = format.lower()  # Not sure though about allowing case-insensitive values.
    if format not in ALLOWED_CONTENT_TYPES:
            raise exceptions.ValidationError(
                f'"{format}" is not a valid format type.')
    return format
```

Also it's a bit confusing that constant is called `ALLOWED_CONTENT_TYPES` but method gets `format_type`

cutwater: Also it would've been easier to move this logic to SearchRequestSerializer
ChoiceField with default value will do.

Issue #1866, opened 2019-05-30
Search by Platform filter causes gateway timeouts from severe performance deficiencies


ironfroggy:
##### SUMMARY
The Platform filter for searches on roles performs significantly worse than other filters, like filtering by Tag or Cloud Platform. I can add many tags to a single search with no noticeable impact on the search API response time. A single Platform filter adds significant time, multiple Platform filters progressively are more and more expensive, easily triggering timeouts around 3 *especially* if combined with the Tag or other filters.

This can reliably bring the whole API down, affecting users who are not performing these searches at all.

##### STEPS TO REPRODUCE
Need a role you can use across a number of platforms, and so add multiple platforms to the search (three or more is good) and watch the loading spinner.

##### EXPECTED RESULTS
Platforms and Cloud Platforms are *like* tags and should have similar performance, especially with even fewer of them involved.

##### ACTUAL RESULTS
Slow response time, API timeout, some times timeouts for entirely other users or pages if hit at the same time.

- x1, 9286 ms
- x2, 10650 ms
- x3, 16146 ms
- x4, 18574 ms
- x5, 19082 ms

No, its not really common for a user to use this many filters, but it demonstrates the increasing cost and keep in mind the performance on production is much worse. When looking to see if it was slow on production already or was a new regression, I accidentally broke production for a few seconds. Even without causing timeouts or crashing, its a very bad experience.

ironfroggy:
Example: searches with 1, 2, or 3 Platforms filtering on.

https://galaxy-qa.ansible.com/search?keywords=&order_by=-relevance&page=1&deprecated=&platforms=Debian

https://galaxy-qa.ansible.com/search?keywords=&order_by=-relevance&page=1&deprecated=&platforms=Debian%20Fedora

https://galaxy-qa.ansible.com/search?keywords=&order_by=-relevance&page=1&deprecated=False&platforms=Debian%20Fedora%20opensuse

PR #1868, opened 2019-05-31
Optimize search performance

cutwater:
Search performance was affected by using `DISTINCT` over all selected
fields and many to many joins.
Many to many joins are replaced with `WHERE .. IN` query, which
improves search query performance 10x to 20x times.

Fixes: #1866

### DIFF
--- a/galaxy/api/internal/search.py
+++ b/galaxy/api/internal/search.py
@@ -69,7 +69,7 @@ class CollectionSearch(BaseSearch):
 
     def count(self):
         """Returns number of matching collections."""
-        return self._base_queryset().count()
+        return self._base_queryset().only('pk').count()
 
     def search(self):
         qs = self._base_queryset().select_related('latest_version')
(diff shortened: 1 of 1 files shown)


galaxy/api/internal/search.py
```diff
@@ -247,24 +248,25 @@ def _base_queryset(self):
...
+                platforms__name__in=platforms)
+            qs = qs.filter(pk__in=platforms_qs)
 
         clouds = self.filters.get('cloud_platforms')
         if clouds:
-            clouds_qs = models.CloudPlatform.objects.filter(
-                name__in=clouds)
-            qs = qs.distinct().filter(cloud_platforms__in=clouds_qs)
+            cloud_qs = models.Content.objects.only('pk').filter(
+                cloud_platforms__name__in=platforms)
```
alikins: Travis tests fail on this, but guessing this should be 'cloud_platforms__name__in=clouds'

cutwater: Thanks. Fixed now.

galaxy/api/internal/search.py
```diff
@@ -247,24 +248,25 @@ def _base_queryset(self):
...
+                platforms__name__in=platforms)
+            qs = qs.filter(pk__in=platforms_qs)
 
         clouds = self.filters.get('cloud_platforms')
         if clouds:
-            clouds_qs = models.CloudPlatform.objects.filter(
-                name__in=clouds)
-            qs = qs.distinct().filter(cloud_platforms__in=clouds_qs)
+            cloud_qs = models.Content.objects.only('pk').filter(
+                cloud_platforms__name__in=clouds)
```
ironfroggy: is this supposed to be `clouds` and not `platforms`?

ironfroggy: oh! already got it.

Issue #1876, opened 2019-06-03
Legacy search API performance deficiencies


cutwater:
##### SUMMARY
Legacy search API is affected by #1866.
While legacy API is not used on UI side, it is still accessed by external clients affecting service performance.

##### EXPECTED RESULTS
Fix #1868 is applied to legacy search API.

PR #1878, opened 2019-06-03
Fix inefficient queries in legacy search API

cutwater:
Improve legacy search API performance by replacing
filters against many-to-many relationships from
`SELECT DISTINCT` and `JOIN` query
to `WHERE .. IN` query.

Fixes: #1876

### DIFF
--- a/galaxy/api/views/search.py
+++ b/galaxy/api/views/search.py
@@ -77,7 +77,7 @@ class ContentSearchView(base.ListAPIView):
 
     def get_queryset(self):
         return (
-            models.Content.objects.distinct()
+            models.Content.objects
             .select_related(
                 'content_type',
                 'namespace',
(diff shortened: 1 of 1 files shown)


galaxy/api/views/search.py
```diff
@@ -236,16 +237,17 @@ def add_namespaces_filter(queryset, namespaces):
     def add_platforms_filter(queryset, platforms):
         if not platforms:
             return queryset
-        return queryset.filter(
-            platforms__in=models.Platform.objects.filter(name__in=platforms))
+        platforms_qs = models.Content.objects.only('pk').filter(
+            platforms__name__in=platforms)
+        return queryset.filter(pk__in=platforms_qs)
```
ironfroggy: Is writing it this appreciably different in the resulting SQL from

```return queryset.filter(platforms__name__in=platforms)```

Comment (cutwater) [reply galaxy/api/views/search.py]: This call ```queryset.filter(platforms__name__in=platforms)``` generates SQL JOIN query, for example:

```
SELECT ...
FROM main_content
JOIN main_content_platforms ON ...
JOIN main_platforms ON ...
WHERE main_platforms.name IN (...)
```
Such JOIN query on many-to-many relationship creates row duplicates, that then can be eliminated by DISTINCT \ DISTINCT ON. Applying dumb `.distinct()` call on a queryset leads to generating `SELECT DISTINCT` over all fields in query, which for this select query can be large (considering selecting fields from related tables).

One possible solution is to select records and join with subquery that returns unique content ids, but Django ORM doesn't provide an obvious interface for custom joins.

Another solution, we use in search, is making `WHERE 'id' IN (SELECT ...)` query, which has similar performance to JOIN with unique IDs.

Issue #1879, opened 2019-06-03
Account/Namespace Problem


mariussturm:
##### SUMMARY

Hi,
I own the account Graylog2 with the namespace Graylog. Unfortunately the account behaves a bit broken. When I search for 'graylog' I get my role (graylog-ansible-role) as first result. But when I click on it, it leads to a 404 page. (Hint: Wrong link location ` https://galaxy.ansible.com/graylog2/graylog-ansible-role` should be `https://galaxy.ansible.com/Graylog2/graylog-ansible-role`)

The other problem is that I can't save the preferences for this account (in order to add/remove owners) it always says: " Error! This provider namespace is already associated with a Galaxy namespace (Graylog2)"

##### STEPS TO REPRODUCE

1. Problem:  Search for 'graylog', click on first match

2. Problem: Click: My content -> Graylog2 -> Edit Properties -> Save

##### ACTUAL RESULTS

```

```

newswangerd:
Hi @mariussturm 

Problem 1 is definitely a bug in the latest version of Galaxy. It looks like we're not looking up content names case insensitively. I'll submit a PR to fix that.

Problem 2 is a little more worrisome. It looks like we somehow ended up with two graylog namespaces: Graylog2 and graylog2 (https://galaxy.ansible.com/api/v1/namespaces/?or__name__icontains=graylog). This shouldn't happen because namespace and role names are supposed to be case insensitively unique.

You can't make any changes to Graylog2 because graylog2 has a provider namespace that has the same name as Graylog2. I can fix this problem by deleting the graylog2 namespace. graylog2 looks like it has a duplicated of `Graylog2.graylog-ansible-role` (https://galaxy.ansible.com/graylog2/graylog) which will get removed if I get rid of the graylog2 namespace. With that in mind, is it okay for me to delete the graylog2 namespace?

mariussturm:
@newswangerd Thanks for the fast response!
Yes, I think you can delete that one, it only holds an older version of the role (1.3.1 vs. 3.0.1).

newswangerd:
@mariussturm I got the old namespace removed. The link on the search page is still broken, but that should get fixed once we can deploy #1880

newswangerd:
We deployed galaxy 3.2.1 this morning with a patch that fixes the link issues, so your links should work again.

mariussturm:
@newswangerd wonderful! Thanks alot for the quick help!

PR #1880, opened 2019-06-03
Make combined API perform case insensitive lookups.

newswangerd:
Fixes: #1879


galaxy/api/internal/views/combined.py
```diff
@@ -48,11 +48,11 @@ def get(self, request):
                 detail='The namespace parameter is required')
 
         collection_filters = {
-            'namespace__name': namespace,
+            'namespace__name__iexact': namespace,
```
ironfroggy: Do we need this change since we _are_ enforcing collections only in lowercase-compliant namespaces?

PR #1888, opened 2019-06-04
Check collection name conflict with multi-content repo name

awcrosby:
This PR prevents importing a collection with a name that conflicts with a multi-content repository name. This keeps the namespace.name unique within galaxy. This warning states that the multi-content repository is deprecated.

### DIFF
--- a/galaxy/api/v2/views/collection.py
+++ b/galaxy/api/v2/views/collection.py
@@ -103,6 +103,7 @@ class CollectionListView(base.APIView):
         namespace = self._get_namespace(data)
         self._check_namespace_access(namespace, request.user)
         self._check_role_name_conflict(namespace, filename.name)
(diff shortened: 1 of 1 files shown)


galaxy/api/v2/views/collection.py
```diff
@@ -165,6 +166,21 @@ def _check_role_name_conflict(self, ns, name):
             'or rename the role via the meta/main.yml role_name attribute'
         )
 
+    def _check_multi_repo_name_conflict(self, ns, name):
+        multi_content_repos = models.Repository.objects.filter(
+            format='multi',
+            name__iexact=name,
+        )
```
ironfroggy: it doesn't look like this filters on the namespace at all

awcrosby: Updated

Issue #1616, opened 2019-03-04
Better distribution search / informations


rockandska:
### Use Case
Not sure if i'm the only one to be not a huge fan of the new Galaxy UI, but i found that the new UI have complicated at least one thing: found a role compatible with multiple distrib

In the old fashion UI, we had an overview directly in the search results of the distribution compatible with the role and it was really handy.

https://d255esdrn735hr.cloudfront.net/graphics/9781788394512/graphics/5be52818-d9be-4e7c-a557-3902983ea559.png

Now, there is no information who let the users know with which distributions is compatible the role, and the filter for distribution only allow "or" operand, so if you looking for a role compatible with "x" and "y" distribution, you're stuck with the need to click on roles to see if they are cross compatible or not and is really annoying.

https://user-images.githubusercontent.com/24546770/53769554-8533c300-3edc-11e9-8e84-c92e7492c95c.png

### Proposed Solution
- add the distributions listed in the meta/main.yml in the search results
- remove the distribution filter by tag and replace it with a old fashion multi-select list to be able to filter "Debian" **and** "EL" compatibles roles.

Regards,

PR #1898, opened 2019-06-06
Set platforms, clouds and tags to AND on search

newswangerd:
fixes: #1616


ironfroggy:
please notify me if this ever merges, so i can adjust testing

PR #1900, opened 2019-06-06
Admin interface for collections and namespaces

cutwater:
* Add simple admin interface for namespaces
* Add simple admin interface for collections
* Fix admin interface for repository versions

### DIFF
--- a/galaxy/main/models/collection.py
+++ b/galaxy/main/models/collection.py
@@ -59,7 +59,7 @@ class Collection(mixins.TimestampsMixin, models.Model):
     # References
     latest_version = models.ForeignKey(
         'CollectionVersion',
-        on_delete=models.PROTECT,
+        on_delete=models.SET_NULL,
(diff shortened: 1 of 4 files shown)


galaxy/main/models/collection.py
```diff
@@ -59,7 +59,7 @@ class Collection(mixins.TimestampsMixin, models.Model):
     # References
     latest_version = models.ForeignKey(
         'CollectionVersion',
-        on_delete=models.PROTECT,
+        on_delete=models.SET_NULL,
```
ironfroggy: This affects delete of the `CollectionVersion`, but you don't add that as a registered model for the admin in this PR.

cutwater: There is circular dependency between Collection and CollectionVersion.
You cannot delete `Collection` if it has `latest_version` field value other than null.

Issue #1905, opened 2019-06-07
Collections: web-site keeps loading


robertdebock:
##### SUMMARY
After publishing [my first collection](https://github.com/robertdebock/ansible-development-environment) the [galaxy page](https://galaxy.ansible.com/robertdebock/ansible_development_environment) keeps loading but never finished.

##### STEPS TO REPRODUCE
Upload [my first collection](https://github.com/robertdebock/ansible-development-environment) and open the related [galaxy page](https://galaxy.ansible.com/robertdebock/ansible_development_environment).

##### EXPECTED RESULTS
...

##### ACTUAL RESULTS
A loading "chain of balls" spinning.

newswangerd:
@robertdebock Try setting a description in your `galaxy.yml`. It looks like there's a bug where the UI crashes if there's no description set.

robertdebock:
Cool, that fixes my issue, thanks!

PR #1906, opened 2019-06-07
Fix collection detail breaking when no description is set.

newswangerd:
Fixes: #1905

### DIFF
--- a/galaxyui/src/app/content-detail/content-header/content-header.component.ts
+++ b/galaxyui/src/app/content-detail/content-header/content-header.component.ts
@@ -137,7 +137,8 @@ export class ContentHeaderComponent implements OnInit {
             iconClass: 'pficon-repository',
             tooltip: 'Collection',
             name: this.collection.name,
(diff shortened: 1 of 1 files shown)


galaxyui/src/app/content-detail/content-header/content-header.component.ts
```diff
@@ -137,7 +137,8 @@ export class ContentHeaderComponent implements OnInit {
             iconClass: 'pficon-repository',
             tooltip: 'Collection',
             name: this.collection.name,
-            description: this.collection.latest_version.metadata.description,
+            description:
+                this.collection.latest_version.metadata.description || '',
```
ironfroggy: Is this the right fix, or should it have been an import error for the metadata to exclude a description in the first place? I did not think it was supposed to be optional.

newswangerd: I looked into this and the description isn't required in the spec, so the importer is currently doing the right thing by ignoring it.

PR #1909, opened 2019-06-10
add link back to Galaxy from docs

samccann:
PR #1374 removed the link that allows a user to go from Galaxy docs back to the Galaxy itself. This PR:

- removes 'documentation' from next to the galaxy logo and places in within the teal Galaxy documentation navigation window.
- Sets the url  on the Galaxy logo to go back to galaxy vs docs.  

BEFORE:

image

AFTER:
image

Note that line 154 keeps the documentation (docs) link within the teal navigation window so the original issue is still fixed with this patch.

### DIFF
--- a/docs/docsite/_themes/sphinx_rtd_theme/layout.html
+++ b/docs/docsite/_themes/sphinx_rtd_theme/layout.html
@@ -130,14 +130,11 @@
         </ul>
     </div>
 
-  <a class="DocSite-nav" href="/docs" style="padding-bottom: 30px;">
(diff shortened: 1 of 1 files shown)


ironfroggy:
@newswangerd can you comment on the original change and its intention?

newswangerd:
> @newswangerd can you comment on the original change and its intention?

I think the original idea behind #1374 was that it was confusing for users that clicking on "Documentation" took them to galaxy. I think the proposed changes here are probably the right way to fix this problem because it allows users to be able to go back to the main site without confusing them with the Documenation text in the logo.

PR #1919, opened 2019-06-12
Prevent duplicate namespace on login when github username case changes

### LINKED ISSUE #1914: Duplicate namespace created when github username case changes
##### SUMMARY
- If a galaxy user’s github username changes case, then during login a new provider namespace is created and associated with the user’s galaxy account https://github.com/ansible/galaxy/blob/devel/galaxy/main/signals/handlers.py#L47
- At this point in time, if the galaxy user’s namespace was not already lower case, a duplicate namespace is created in lower case https://github.com/ansible/galaxy/blob/devel/galaxy/main/signals/handlers.py#L69

##### STEPS TO REPRODUCE

* Must have galaxy namespace of mixed case
* Change github username case in github
* Log in to galaxy
* A second namespace that is all lowercase will show up associated to your galaxy user

##### EXPECTED RESULTS
No second namespace is created

awcrosby:
Fixes #1914

### DIFF
--- a/galaxy/main/signals/handlers.py
+++ b/galaxy/main/signals/handlers.py
@@ -43,11 +43,8 @@ def user_logged_in_handler(request, user, **kwargs):
     username = social.extra_data.get('login')
     sanitized_username = username.lower().replace('-', '_')
 
-    try:
-        models.ProviderNamespace.objects.get(name=username)
+    if models.ProviderNamespace.objects.filter(name=username).exists():
         return
-    except models.ProviderNamespace.DoesNotExist:
-        pass
 
     # User is not associated with any Provider Namespaces, so we'll attempt
     # to create one, along with associated Namespace.
@@ -56,7 +53,6 @@ def user_logged_in_handler(request, user, **kwargs):
     name = social.extra_data.get('name') or username
 
     defaults = {
-        'name': sanitized_username,
         'description': name,
         'avatar_url': social.extra_data.get('avatar_url'),
         'location': social.extra_data.get('location'),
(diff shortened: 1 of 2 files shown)


galaxy/main/signals/handlers.py
```diff
@@ -44,35 +44,39 @@ def user_logged_in_handler(request, user, **kwargs):
...
-    # User is not associated with any Namespaces, so we'll attempt
-    # to create one, along with associated Provider Namespaces.
-    provider = models.Provider.objects.get(name__iexact="github")
+    # User is not associated with any Provider Namespaces, so we'll attempt
+    # to create one, along with associated Namespace.
 
     # if name doesn't exist, set it to login
     name = social.extra_data.get('name') or username
 
     defaults = {
```
cutwater: If you're reusing defaults dictionary, it's not the best approach to add query-specific params, then execute a query and then remove those specific params in order to execute another one.

Please use common dictionary and add query-specific params when they're required:


```
defaults = {
    ...
}

ns_defaults = {'name': sanitized_username, **defaults}
models.Namespace.objects.get_or_create(..., defaults=ns_defaults)

File: galaxy/main/signals/handlers.py [hunk 2]
```diff
@@ -43,15 +43,11 @@ def user_logged_in_handler(request, user, **kwargs):
     username = social.extra_data.get('login')
     sanitized_username = username.lower().replace('-', '_')
 
-    try:
-        namespace = models.ProviderNamespace.objects.get(name=username)
+    if models.ProviderNamespace.objects.filter(name=username).exists():
```
newswangerd: Won't this create a duplicate provider namespace if the user changes their username or the capitalization on their username? If so, is that a problem?

galaxy/main/signals/handlers.py
```diff
@@ -64,18 +60,26 @@ def user_logged_in_handler(request, user, **kwargs):
...
-        namespace.owners.add(user)
-    defaults['description'] = social.extra_data.get('bio') or name
-    defaults['followers'] = social.extra_data.get('followers')
-    defaults['display_name'] = name
-    defaults['avatar_url'] = social.extra_data.get('avatar_url')
+
+    ns_defaults = {'name': sanitized_username, **defaults}
+
+    # Create lowercase namespace if case insensitive get does not find match
+    namespace, _ = models.Namespace.objects.get_or_create(
```
newswangerd: `.get` throws an error if two items match the search. If `.get_or_create` works the same way, will this cause an exception if there are two namespaces with different capitalization?

galaxy/main/signals/handlers.py
awcrosby: On these changes yes a new provider namespace is created, I don't think that is an issue as we have been operating that way, and this way our system has a link to what the changed github username is via the provider namespace, since our user.username will no longer match the github username.

newswangerd:
So, one thing I noticed is that no matter how much I change my github username, my galaxy username stays the same:

Screen Shot 2019-06-12 at 4 17 50 PM

I also noticed that when I change my github username, it creates a new namespace with the new GH username:

Screen Shot 2019-06-12 at 4 18 53 PM

Is there any way that we could make it so that your namespace is just a sanitized version of your ansible username instead of whatever the GitHub API throws at us when the user logs in? I'm thinking that if a user changes their username, it should just add a new provider namespace to their existing namespace instead of creating a new namespace.

awcrosby:
@newswangerd while that sounds like a potentially good approach, I would be hesitant to make changes as I'm not sure there is a specific problem we are running in to with users, and making a change could cause problems, and the namespace matching github as it does may be by-design.

galaxy/main/signals/handlers.py
awcrosby: I hadn't tested when there already was a duplicated namespace, I will check and fix if needed. Thanks!

Issue #1920, opened 2019-06-13
Renaming github namespace from: uniQconsulting -> uniqconsulting


joe-speedboat:
##### SUMMARY
Hi 
We recently cleaned up our namespace again to align with github and role names:
To make the import and naming more robust, we changed uQc account from:
uniQconsulting to uniqconsulting
Now problem that I cannot access my roles anymore from within Galaxy Webpage:
https://galaxy.ansible.com/uniqconsulting (is showing all the roles)
If i click on one role, I get a "Page not found" error message.

##### STEPS TO REPRODUCE
Now problem that I cannot access my roles anymore from within Galaxy Webpage:
https://galaxy.ansible.com/uniqconsulting (is showing all the roles)
If i click on one role, I get a "Page not found" error message.

##### EXPECTED RESULTS
Now problem that I cannot access my roles anymore from within Galaxy Webpage:
https://galaxy.ansible.com/uniqconsulting (is showing all the roles)
If i click on one role, I would expect to see role details.
##### ACTUAL RESULTS
Now problem that I cannot access my roles anymore from within Galaxy Webpage:
https://galaxy.ansible.com/uniqconsulting (is showing all the roles)
If i click on one role, I get a "Page not found" error message.
##### QUESTION
May you please purge all the roles from https://galaxy.ansible.com/uniqconsulting so I can run a fresh import that hopefully will get accepted by webinterface of galaxy

Thanks a lot
Chris

awcrosby:
@joe-speedboat - those 11 roles have been deleted, let us know if you have issues after you re-import them.
Thanks!

uniqconsulting:
@awcrosby 
I reimported and still get the same "Page not found" message if I try to view role details from within ansible galaxy page:
eg: https://galaxy.ansible.com/uniqconsulting/graylog3

btw, cli still works as expected (no issues there!?):
```
[chris@sp-remote-01 files]$ ansible-galaxy search uniqconsulting

Found 10 roles matching your search:

 Name                            Description
 ----                            -----------
 uniqconsulting.collabora        installation and configuration of collabora CODE with nginx reverse proxy
 uniqconsulting.graylog3         Install graylog3 and configure single server instance
 uniqconsulting.guacamole        Install and configure Guacamole with Nginx ssl reverse proxy
 uniqconsulting.mariadb          installation and configuration of mariadb (mysql)
 uniqconsulting.network_setup    Network Setup
 uniqconsulting.nextcloud        nextcloud installation and configuration
[...]

PR #1928, opened 2019-06-18
Fix lookup of duplicate repo namespaces on combined view

awcrosby:
Corrects server error `MultipleObjectsReturned at /api/internal/ui/repo-or-collection-detail/` when there are multiple namespace names under different cases:
`/api/internal/ui/repo-or-collection-detail/?namespace=[namespace_name]&name=[repo_role_name]`

Fixes: #1920


newswangerd:
We should probably fix this by removing the duplicate namespaces, not by updating the API to work with invalid data.

awcrosby:
Closing as user's issue was resolved via removing the duplicate namespace.

PR #1929, opened 2019-06-20
broken URL

thedoubl3j:
broken url that was not resolving due to ansible misspelling.  
+label: docsite_pr


chouseknecht:
Thanks for fixing this!

PR #1972, opened 2019-08-02
Added support for Github Enterprise instances

egmar:
With this pull request I'm proposing change which would add support for Github Enterprise instances.
Main set of changes focuses on removing hard-coded references to `github.com` and ensure correct address is return by API for `ansible-galaxy` CLI utility.

A patch for ansible/ansible will be submitted later to add support for GHE as well.

Signed-off-by: Egor Margineanu <egor.margineanu@gmail.com>

### DIFF
--- a/galaxy/api/githubapi.py
+++ b/galaxy/api/githubapi.py
@@ -59,7 +59,8 @@ class GithubAPI(object):
                 "User does not have a GitHub OAuth token"
             )
         try:
-            client = Github(base_url=settings.GITHUB_SERVER, login_or_token=gh_token.token)
+            client = Github(base_url=settings.GITHUB_SERVER,
+                            login_or_token=gh_token.token)
         except GithubException as exc:
             raise Exception("Failed to connect to the GitHub API {0} - {1}"
                             .format(exc.data, exc.status))
--- a/galaxy/api/githubapi.py
+++ b/galaxy/api/githubapi.py
@@ -23,6 +23,7 @@ from github import Github
 from github.GithubException import GithubException
 
 from allauth.socialaccount.models import SocialToken
+from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from galaxy.main.models import Provider
 
@@ -58,7 +59,7 @@ class GithubAPI(object):
                 "User does not have a GitHub OAuth token"
             )
         try:
-            client = Github(gh_token.token)
+            client = Github(base_url=settings.GITHUB_SERVER, login_or_token=gh_token.token)
         except GithubException as exc:
             raise Exception("Failed to connect to the GitHub API {0} - {1}"
                             .format(exc.data, exc.status))
--- a/galaxy/api/githubapi.py
+++ b/galaxy/api/githubapi.py
@@ -23,8 +23,8 @@ from github import Github
 from github.GithubException import GithubException
 
 from allauth.socialaccount.models import SocialToken
-from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+from galaxy.common.github import get_github_api_url
 from galaxy.main.models import Provider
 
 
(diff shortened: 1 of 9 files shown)


chouseknecht:
@cutwater 

Before we merge this. It seems we're defining a GitHub server 2 places in our settings. One is `settings.SOCIALACCOUNT_PROVIDERS['github'].GITHUB_URL` and the other is `settings.GITHUB_SERVER`. Wondering if we can standardize on one, likely the first one.

egmar:
I believe these are two distinct endpoints. First, per my understanding, is part of Django OAuth mechanisms. Second is the URI to Github API endpoint and it might differ. For public Github it is `https://api.github.com`, for Github Enterprise instances it is usually `https://github.domain.com/api/v3`, though one could expect it to be different, e.g.: `https://api.github.domain.com`

galaxy/api/serializers/repository.py
```diff
@@ -166,7 +167,15 @@ def get_summary_fields(self, instance):
...
         server = ''
         if instance.provider_namespace.provider.name.lower() == 'github':
-            server = 'https://github.com'
+            if 'github' in settings.SOCIALACCOUNT_PROVIDERS:
+                if 'GITHUB_URL' in settings.SOCIALACCOUNT_PROVIDERS['github']:
+                    server = settings \
+                        .SOCIALACCOUNT_PROVIDERS['github'] \
+                        .GITHUB_URL
+                else:
+                    server = 'https://github.com'
```
cutwater: Please remove duplicated `else` clauses

galaxy/main/models/repository.py
```diff
@@ -106,7 +106,18 @@ class Meta:
...
     @property
     def clone_url(self):
-        return "https://github.com/{user}/{repo}.git".format(
+        server = ''
+        if 'github' in settings.SOCIALACCOUNT_PROVIDERS:
+            if 'GITHUB_URL' in settings.SOCIALACCOUNT_PROVIDERS['github']:
+                server = settings \
+                    .SOCIALACCOUNT_PROVIDERS['github'] \
+                    .GITHUB_URL
+            else:
```
cutwater: Same here. Also this code seems to be duplicate of code above.

galaxy/main/models/repository.py
```diff
@@ -119,6 +130,20 @@ def github_user(self):
     def github_repo(self):
         return self.original_name
 
+    @property
+    def github_server(self):
+        server = ''
+        if 'github' in settings.SOCIALACCOUNT_PROVIDERS:
```
cutwater: Same here.

galaxy/api/githubapi.py
```diff
@@ -58,7 +59,8 @@ def get_client(self):
                 "User does not have a GitHub OAuth token"
             )
         try:
-            client = Github(gh_token.token)
+            client = Github(base_url=settings.GITHUB_SERVER,
```
cutwater: To avoid confusion with `SOCIALACCOUNT_PROVIDERS['github'].GITHUB_URL` would it be better to indicate that this  url is related to API explicitly (e.g. `GITHUB_API_URL`)

egmar: Understand, indeed it would make sense, though this would require to validate if `GITHUB_SERVER` is not used in other parts of the code. 
On Github Enterprise topic, I'm going to look into https://python-social-auth.readthedocs.io/en/latest/backends/github_enterprise.html and see if it would cover some of the changes I've included in this PR. Specifically, API is one of the required parameters.

egmar: Yes, it would make sense. Though I'm not sure if `GITHUB_SERVER` parameter isn't used in other parts of the code.

galaxy/main/models/repository.py
egmar: Fixed in last commit

galaxy/main/models/repository.py
egmar: Fixed in last commit

galaxy/api/githubapi.py
egmar: I've looked how it's done in **django-allauth** library (see https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/github/views.py ) and it looks like `GITHUB_SERVER` parameter in settings.py is not used at all. I implemented similar way in `galaxy\common\github.py`.

arielrolfo:
hello guys, I'm quite interested on this feature, is there any plan to merge it any time soon ?

chouseknecht:
Merging for further testing in QA.

PR #2009, opened 2019-09-06
Use relative paths for 'available_versions' paths

alikins:
This is to allow the ApiRootView to be somewhere
other than /api.

Needed for ansible-galaxy compat with galaxy v1/v2 and automation hub v3.
(ah hub api root is /api/automation-hub/ not /api/)

### DIFF
--- a/galaxy/api/views/views.py
+++ b/galaxy/api/views/views.py
@@ -115,13 +115,12 @@ class ApiRootView(base_views.APIView):
 
     def get(self, request, format=None):
         # list supported API versions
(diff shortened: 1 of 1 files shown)


ironfroggy:
Unfortunately, this seems to immediately break all current versions of `ansible-galaxy`, which are not properly resolving this relative URL. It is being resolved relative to the hostname, rather than the current endpoint. This results in attempting to fetch, for example, an authentication token from `https://galaxy.ansible.com/v1/tokens/` erroneously, resulting in a 404. This change makes sense, but is not backwards compatible with any current version of the client.

Issue #2006, opened 2019-09-04
Require collection to have 'repository' or allow collection to be browsed via Galaxy UI


geerlingguy:
### Use Case

Ansible Galaxy roles had to be tied to a GitHub repository. Being tied to one specific service was annoying in many cases, but one benefit of that was you could always and quickly browse the source of the role, without having to download it to your local machine.

Ansible Galaxy collections do _not_ have to be tied to a GitHub repository (the `repository` field in `galaxy.yml` is optional). Nor do they need to be tied to any kind of public repository (meaning collections can technically be closed source).

This creates two problems, though:

  1. There is no way to see what's inside a collection (e.g. https://galaxy.ansible.com/devoperate/base) without downloading it locally.
  2. Every time someone does wish to evaluate the code in a collection, they must download it, thus inflating the download count (one of the metrics which may or may not be important in the search rankings of a collection).

### Proposed Solution

Require the `repository` field in `galaxy.yml`.

(Yes, I know that technically someone could build a collection from anywhere as long as they have credentials, so the source should not be considered as canonical, either—but for 99% of collections, the source would have a documented build process that allows me to also verify that the collection I'm getting did not suffer from a code injection attack during its build on some crusty old internal Jenkins server... basically I want to be able to have reproducible builds for collections.)

### Alternatives

Another option would be to build some sort of mechanism that allows browsing the contents of a collection in the Galaxy UI. There be dragons; this is why it's nice to be able to link back to Gogs, Gitea, GitLab, GitHub, etc. (see proposed solution).

### Implementation

N/A

chouseknecht:
According to the Collection Spec doc there is a required link to the source repository in `galaxy.yml`. Not sure if we're enforcing this during the import process, but seems we should be, and we should surface it in the UI as a link.

newswangerd:
If the link is there it will always show up in the UI. We just need to enforce that the link is there in the importer.

geerlingguy:
@chouseknecht / @newswangerd - Currently, in the documentation, `repository` is listed as an _optional_ field:

<img width="446" alt="Screen Shot 2019-09-23 at 10 43 27 AM" src="https://user-images.githubusercontent.com/481677/65435769-0ad27e00-ddef-11e9-981f-920fa08e2a1b.png">

thaumos:
What would we do in the case of there not being an open repo existing? This is the one thing I foresee causing this to be a cause for issue.

awcrosby:
[...]

PR #2049, opened 2019-10-01
Require collection metadata repository be non-null

awcrosby:
Resolves: #2006


pabelanger:
Documentation was not properly updated: https://docs.ansible.com/ansible/devel/dev_guide/collections_galaxy_meta.html#collections-galaxy-meta

repository is still listed as option, but galaxy now rejects a collection with it missing.

PR #2113, opened 2019-11-19
[docs/installing] Replace nonexistent example 'debops.apt'

fhemberger:
Looks like the mentioned example 'debops.apt' doesn't exist any more. Replaced it with 'debops.debops'. 🤷‍♂ 

Signed-off-by: Frederic Hemberger <mail@frederic-hemberger.de>


drybjed:
Although, `debops.debops` is now an Ansible Galaxy Collection, not a single role. Since this is an example for roles, you should probably switch to something different, perhaps one of @geerlingguy roles?

chouseknecht:
Closing, as the suggested fix is a Collection and not a role.

fhemberger:
Would something simple like `geerlingguy.ntp` be okay?

Issue #2170, opened 2019-12-21
API timeouts


till:
##### SUMMARY

Cannot install roles from Ansible Galaxy.

##### STEPS TO REPRODUCE

 1. Add roles to requirements.yml
 2. Run galaxy install

##### EXPECTED RESULTS

 1. Roles are successfully installed

##### ACTUAL RESULTS

```
- extracting cloudalchemy.alertmanager to /root/.ansible/roles/cloudalchemy.alertmanager
--
647 | - cloudalchemy.alertmanager (0.13.5) was installed successfully
648 | [WARNING]: - cloudalchemy.grafana was NOT installed successfully: Failed to
649 | get data from the API server (https://galaxy.ansible.com/api/): The read
650 | operation timed out
651 | ERROR! - you can use --ignore-errors to skip failed roles and finish processing the list.
```

This is from our CI system, when I re-run my builds, the error moves on to a different role.

till:
Btw, a proper status page would be neat. :)

mashayev:
Same issue for me.
As @till said ->  A status page would be nice!

Flarf:
Same here.  Discovered the issue around 10PM MST, and still seeing the issue.

chouseknecht:
Galaxy is experiencing some intermittent problems. We’re looking into it. Will provide an update when we know more. Thanks to those that reported seeing issues.

nqb:
Issue seems still present.

mashayev:
@chouseknecht Any ETA ?

mashayev:
As a workaround(I don't like it, but at least it works) you can add a for loop and limit it to 10 or something:
`for i in {0..10}; do ansible-galaxy install -p <blah> <blah> && break; done`

This is a temp fix till Galaxy will be fixed.

till:
My workaround is to download roles from git/github directly:

```diff
 - name: robertdebock.bootstrap
+  src: https://github.com/robertdebock/ansible-role-bootstrap
   version: 3.0.4	  version: 3.0.4
```

chouseknecht:
Influx database where we capture monitoring data grew passed available volume space. Working on cleaning it up. Should be fixed within the next hour or two.

chouseknecht:
Fixed. Many thanks to @cutwater

ssbarnea:
Any regression measures implemented so we do not see this happening again, for same or similar reasons? We should have a live status page and alerts, not to rely on users reporting it on irc first.

[...]

PR #2172, opened 2019-12-22
Feature: monitor endpoint for galaxy

till:
Endpoint: /api/monitor/

Checks:

 - influxdb connection
 - rabbitmq connection
 - postgresql connection (and migration state)
 - redis connection

---

image

---

Motivation: I couldn't find one, and thought this would be a good idea to have.
You can stick this into a monitoring tool and look for "error" and hopefully get
someone on it, before Galaxy is brittle for over a day.

Related: ansible/galaxy#2170

---

I ran all tests locally, and formatters and prettifiers. Kudos for the docs and the setup.

### DIFF
--- a/galaxy/api/views/views.py
+++ b/galaxy/api/views/views.py
@@ -51,6 +51,13 @@ from galaxy.main.celerytasks import tasks as celerytasks
 from galaxy.main import models
 from galaxy.common import version, sanitize_content_name
 
+# monitor
+from kombu import Connection as RabbitMQ
+from django.db.migrations.executor import MigrationExecutor
+from django.db import connections, DatabaseError, DEFAULT_DB_ALIAS
+import influxdb
+import redis
+
 
 logger = logging.getLogger(__name__)
 
@@ -66,6 +73,7 @@ __all__ = [
     'ImportTaskLatestList',
     'ImportTaskList',
     'ImportTaskNotificationList',
+    'MonitorRootView',
     'PlatformDetail',
     'PlatformList',
     'ProviderRootView',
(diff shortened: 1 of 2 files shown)


till:
@chouseknecht feedback?

gundalow:
If we go this route, it's easy for me to add this into uptimerobot which alerts in the monitoring slack channel.

galaxy/api/views/views.py
```diff
@@ -51,6 +51,13 @@
 from galaxy.main import models
 from galaxy.common import version, sanitize_content_name
 
+# monitor
```
cutwater: This module is a leftover of galaxy before refactoring. New views should not be added here.
Can you please move your code to a separate module (e.g. `galaxy/api/views/monitor.py`)?

galaxy/api/views/views.py
```diff
@@ -51,6 +51,13 @@
...
+from kombu import Connection as RabbitMQ
```
cutwater: Please organize imports according to [PEP8](https://www.python.org/dev/peps/pep-0008/#imports):

> 1. Standard library imports.
> 2. Related third party imports.
> 3. Local application/library specific imports.

galaxy/api/views/views.py
```diff
@@ -109,6 +117,71 @@ def filter_rating_queryset(qs):
 # -----------------------------------------------------------------------------
 
 
+class MonitorRootView(base_views.APIView):
+    """ Monitor resources """
+    permission_classes = (AllowAny,)
+
+    monitor_response = OrderedDict()
```
cutwater: Why `monitor_response` is a class variable?

galaxy/api/views/views.py
```diff
@@ -109,6 +117,71 @@ def filter_rating_queryset(qs):
...
+                password=settings.INFLUX_DB_PASSWORD
+            )
+
+            try:
+                influxdb_client.ping()
+                self.monitor_response['influx'] = 'ok'
+            except influxdb.client.InfluxDBClientError:
+                self.monitor_response['influx'] = 'error'
+
+            influxdb_client.close()
```
cutwater: This line may not be called in case of any unexpected exceptions above. Please use context managers (`with` statement), [contextlib.ExitStack](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack) or `try .. finally`.
Because InfluxDBClient does not support context management protocol, ExitStack or `try .. finally` will be appropriate approach for guaranteed closing the connection.

galaxy/api/views/views.py
```diff
@@ -109,6 +117,71 @@ def filter_rating_queryset(qs):
...
+                self.monitor_response['migrations'] = 'needed'
+
+            self.monitor_response['postgresql'] = 'ok'
+        except DatabaseError:
+            self.monitor_response['postgresql'] = 'error'
+
+        redis_client = redis.Redis(
+            host=settings.REDIS_HOST,
+            port=settings.REDIS_PORT,
+            db=0
```
cutwater: Why `db=0` is hardcoded here? It's a default value for `redis.Redis` client.

galaxy/api/views/views.py
```diff
@@ -109,6 +117,71 @@ def filter_rating_queryset(qs):
 # -----------------------------------------------------------------------------
 
 
+class MonitorRootView(base_views.APIView):
+    """ Monitor resources """
+    permission_classes = (AllowAny,)
+
+    monitor_response = OrderedDict()
+
+    def get(self, request, format=None):
```
cutwater: This method includes logic for different operations (e.g. checking influxdb connection, database connection and migrations, redis and rabbitmq connection). Please use decomposition for this logic. It would be nice to split it into different methods that serve one specific purpose.

galaxy/api/views/views.py
```diff
@@ -109,6 +117,71 @@ def filter_rating_queryset(qs):
...
+
+        rabbit_client = RabbitMQ(settings.BROKER_URL)
+
+        rabbit_client.connect()
+        if rabbit_client.connected:
+            self.monitor_response['rabbitmq'] = 'ok'
+        else:
+            self.monitor_response['rabbitmq'] = 'error'
+
+        rabbit_client.close()
```
cutwater: ditto

galaxy/api/views/views.py
```diff
@@ -109,6 +117,71 @@ def filter_rating_queryset(qs):
...
+            host=settings.REDIS_HOST,
+            port=settings.REDIS_PORT,
+            db=0
+        )
+
+        try:
+            redis_client.ping()
+            self.monitor_response['redis'] = 'ok'
+        except redis.exceptions.RedisError:
+            self.monitor_response['redis'] = 'error'
```
cutwater: Why `redis_client` is not closed?

galaxy/api/views/views.py
till: @cutwater I am not familiar, is this about the order it's written?

galaxy/api/views/views.py
till: @cutwater I tried to follow the file, I am not a Python developer. Would you like it not to be a class variable?

galaxy/api/views/views.py
till: I'll remove this line.

galaxy/api/views/views.py
till: I guess I didn't know, or forgot. Should I use the `finally` as well?

galaxy/api/views/views.py
till: `finally`?

till:
@cutwater I responded to everything, I believe. Thank you for your patience. Would you mind a follow-up? Then I'll get started.

galaxy/api/views/views.py
chouseknecht: Yes, it's about the order.

chouseknecht:
Closing and replacing with #2211. Just trying to help out and speed this along...

PR #2229, opened 2020-02-05
Removed redundant code for validating GitHub API availability

egmar:
In GHE (and I assume Github Enterprise Cloud), accessing `https://github.domain/api/v3` without authentication token returns following:

```
{
  "message": "Must authenticate to access this API.",
  "documentation_url": "https://developer.github.com/enterprise/2.17/v3"
}
```

Current code already checks if a token is supplied, and later code for `/user/` API is using it and when API is unavailable is handling such errors. Ideally, I believe message property can be in `ValidationError` if necessary

Signed-off-by: Egor Margineanu <egor.margineanu@gmail.com>

### DIFF
--- a/galaxy/api/views/token.py
+++ b/galaxy/api/views/token.py
@@ -76,14 +76,6 @@ class TokenView(base_views.APIView):
             raise ValidationError({'detail': "Invalid request."})
 
         try:
-            git_status = requests.get(get_github_api_url())
-            git_status.raise_for_status()
-        except Exception:
-            raise ValidationError({
-                'detail': "Error accessing GitHub API. Please try again later."
-            })
-
-        try:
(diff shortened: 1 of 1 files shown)


chouseknecht:
Our primary focus is GitHub cloud. The first test, the one you removed, is making sure the API is actually accessible. Sometimes the cloud API can't be reached, and the resulting error has nothing to do with the user's access token.

egmar:
I understand. 
My main concern here, is `git_status.raise_for_status()` being too wide in its scope. For example, for GHE, the return code is `HTTP/1.1 401 Unauthorized` with JSON message, which in context of the code, if going more granular in regards to specific return codes, can be still considered OK if followed by a request which includes a token.

PR #2231, opened 2020-02-06
Limit maximum requests number by gunicorn worker

cutwater:


cutwater:
This is a workaround for constantly increasing request latency.

PR #2269, opened 2020-02-26
Render tables in README files

### LINKED ISSUE #2245: Render tables in Ansible galaxy
### Use Case
Right now I have multiple Ansible Collections that use their README.md with tables (either markdown or html tables that are fully supported by markdown standards and Github.com)

Here is what Github.com renders->
image

Right now this gets rendered like this in galaxy.ansible.com->
image

### Proposed Solution
Just render tables please

### Alternatives
I have no idea

### Implementation
see above

bmclaughlin:
Before:
2245-before

After:
image

Fixes #2245.


IPvSean:
love it!

PR #2554, opened 2020-11-19
Fix source handling for private repositories

SimonStefan:
Use correct format for private repositories as stated here:
https://github.com/ansible/ansible/issues/8937#issuecomment-55097691

Tested with bitbucket.org

+label: docsite_pr

Signed-off-by: Stefan Simon <stefan.simon@bytepoets.com>


SimonStefan:
I had the version 2.9 of ansible-galaxy installed on the test machine, so this PR should not be relevant anymore

PR #2578, opened 2020-12-15
Fix the check for max size of docs files

awcrosby:
The check for role README max file size during the Community Galaxy
import process uses the wrong calculation to to check for 512 KiB,
imports currently fail when greater than 256 KiB.

AAH-297

### DIFF
--- a/galaxy/importer/utils/readme.py
+++ b/galaxy/importer/utils/readme.py
@@ -34,7 +34,7 @@ README_MIMETYPES = {
     '.md': 'text/markdown',
     '.rst': 'text/x-rst',
 }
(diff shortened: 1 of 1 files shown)


alikins (changes requested):
Add a little more detail to the commit message (and maybe to https://issues.redhat.com/browse/AAH-297).

ie, what the new max README size is. And a note about how the incorrect README_MAX_SIZE manifests itself (ie, does it show an error to users? logs? 500?)

Code looks good.

cutwater (changes requested):
See comment from @alikins . Otherwise LGTM.

alikins (approved):
LGTM

Issue #2622, opened 2021-02-22
php-versions role name was just changed again


geerlingguy:
##### SUMMARY

I'm trying to update my role so it works with ansible-lint 5.x, and that version requires standalone roles to have a valid `role_name` in the `meta/main.yml` file.

So I added that value to my `php-versions` role (which _was_ here: https://galaxy.ansible.com/geerlingguy/php-versions), and imported it on Galaxy.

I saw in the import task output:

```
Updating repository name "php-versions" -> "php_versions" 
```

And indeed, the role has vanished from the old URL and is not installable when requiring `geerlingguy.php-versions` anymore.

I guess I'll have to figure out some workaround to get ansible-lint working, but in the mean time, do you know if the ability to set a `role_name` in the meta/main.yml file and it not blow out the old name be fixed, or should I just not use that feature and try to work with the ansible-lint maintainer to figure out some other way of supporting standalone Galaxy roles?

The role has a quarter-million downloads and is used by thousands of developers in the PHP ecosystem... I really don't want to break their workflows for too long.

##### STEPS TO REPRODUCE

  1. Have a role on Ansible Galaxy with a dash in the name.
  2. Set the name of the role in meta/main.yml
  3. Import the role.

##### EXPECTED RESULTS

The role should not change names replacing dashes with underscores.

##### ACTUAL RESULTS

The old role is wiped out and moved to an entirely new name, and `ansible-galaxy` installs for the old role name completely fail (without even a redirect to the new role name).

Related: https://github.com/ansible/galaxy/issues/2393

newswangerd:
@geerlingguy @awcrosby is working on a fix for this today.

awcrosby:
If a user imports a new standalone role, we want to [enforce](https://github.com/ansible/galaxy/blob/devel/galaxy/importer/loaders/role.py#L387-L388) the usage of underscores not dashes in the role name, so it is easier to move that role to a collection. We also want to support a user’s ability to change a role name by setting the meta/main.yml `role_name` https://galaxy.ansible.com/docs/contributing/importing.html#role-name-limitations:
> To override the default name for an Ansible role, set the role_name attribute in the role meta/main.yml file

But for this particular issue… we shouldn’t break existing roles. So we are planning to add a new logic check to prevent this for existing roles.

Regarding `ansible-lint` requiring the meta/main.yml `role_name`, we may want to reconsider that since we state in docs to set it to override the default role name.

geerlingguy:
[...]

PR #2626, opened 2021-02-25
Prevent role rename if metadata and database role names match

awcrosby:
* Centralize the validation of meta/main.yml `role_name`
* Skip validation if metadata `role_name` already matches database role name

Resolves: https://github.com/ansible/galaxy/issues/2622
JIRA tracker issue: AAH-388


newswangerd (approved):
I haven't tested this, but it looks reasonable.

PR #2632, opened 2021-03-04
Update python dependencies

awcrosby:
* Reformat dev-requirements.txt per latest pip-tools
* Update requirements.in django, galaxy-importer, and remove bleach deps
* pip-tools update requirements files for django and galaxy-importer
* pip-tools update requirements files for aiohttp
* Update dev-requirements.txt with latest cryptography 
* Update code to use bleach_allowlist

### DIFF
--- a/galaxy/importer/utils/readme.py
+++ b/galaxy/importer/utils/readme.py
@@ -22,7 +22,7 @@ import os
 
 import markdown
 import bleach
-from bleach_whitelist import markdown_tags, markdown_attrs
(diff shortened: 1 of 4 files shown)


cutwater (approved):
If this PR is going to be merged as is and not squashed, then it's LGTM.
Normally I'd review a PR as a single diff, and then it'd be impossible to distinct format change of `requirements.txt` and upgrade of `bleach-allowlist`

PR #2677, opened 2021-04-21
AAH-513

alikins:
Issue: AAH-513

### DIFF
--- a/galaxy/main/signals/handlers.py
+++ b/galaxy/main/signals/handlers.py
@@ -44,6 +44,7 @@ def user_logged_in_handler(request, user, **kwargs):
     sanitized_username = username.lower().replace('-', '_')
 
     if models.ProviderNamespace.objects.filter(name=username).exists():
+        logger.debug('Found ProviderNamespace name=username (%s)', username)
         return
 
     # User is not associated with any Provider Namespaces, so we'll attempt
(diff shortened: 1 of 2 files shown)


galaxy/main/signals/handlers.py
```diff
@@ -81,9 +77,29 @@ def user_logged_in_handler(request, user, **kwargs):
     }
     provider = models.Provider.objects.get(name__iexact="github")
 
-    models.ProviderNamespace.objects.get_or_create(
```
chouseknecht, line 84: I don't believe this should be changed. There are old namespaces, prior to Galaxy 3, that didn't convert right. In that case the, I think the .get() will fail.

alikins: Not sure I follow.  Which .get()? 

This line is replaced with either:
``` python
models.ProviderNamespace.objects.get_or_create(name=username,
                                                                                  provider=provider,
                                                                                  defaults=provider_ns_defaults)
```

or 
``` python
models.ProviderNamespace.objects.get_or_create(name=username,
                                                                                  namespace=namespace,
                                                                                  provider=provider,
                                                                                  defaults=provider_ns_defaults)
```

chouseknecht: I see. You moved the get_or_create below. Nvm.

PR #2678, opened 2021-04-21
Add management command to delete CollectionVersions

awcrosby:
Remove CollectionVersion's from Repository and database.

For use when Collection has subset of its CollectionVersion to be removed, if you want to remove all CollectionVersion for a Collection, more analysis should be done. 

Call `bump_repo_version` with list of CollectionVersion ids then after task is complete call `delete_old_repo_versions`.

AAH-445

### DIFF
--- a/galaxy/main/management/commands/delete_collection_version.py
+++ b/galaxy/main/management/commands/delete_collection_version.py
@@ -1,0 +1,83 @@
+from django.core.management.base import BaseCommand
+from dynaconf.contrib import django_dynaconf  # noqa
+
+from pulpcore.app.models import Repository, RepositoryVersion
+from pulpcore.app.tasks import orphan_cleanup
+from pulpcore.app.tasks.repository import add_and_remove
+
+from galaxy.common.tasking import create_task
+
+
+class Command(BaseCommand):
+    help = ("Remove CollectionVersion's from Repository and database. "
+            "For use when Collection has subset of its CollectionVersion to "
+            "be removed, if you want to remove all CollectionVersion for a "
+            "Collection, more analysis should be done. Call bump_repo_version "
+            "with list of CollectionVersion ids then "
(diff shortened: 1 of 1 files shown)


galaxy/main/management/commands/delete_collection_version.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def handle(self, *args, **kwargs):
+        assert Repository.objects.all().count() == 1, 'Expected 1 repository'
+        repo = Repository.objects.get()
+
+        # Resources to lock when running tasks. Need path component of the
+        # resource URI of repository, but 'repositories-detail' view is not
+        # exposed within galaxy and pulpcore.tasking.util.get_url()
+        # reverse lookup causes django.urls.exceptions.NoReverseMatch error.
+        # resources = [repo]
```
awcrosby, line 39: @daviddavis I was unable to get the `Repository` lock working with `pulpcore==3.0.0b21` along with how `galaxy` uses it. If that is the case, we may need to run this in a maintenance window.

galaxy/main/management/commands/delete_collection_version.py
```diff
@@ -0,0 +1,83 @@
+from django.core.management.base import BaseCommand
+from dynaconf.contrib import django_dynaconf  # noqa
```
awcrosby, line 2: @cutwater fyi the loading of settings issue I found when calling this via `django-admin shell` were resolved once I used `python manage.py shell`

galaxy/main/management/commands/delete_collection_version.py
daviddavis: +1 to running this in a maintenance window. You want to avoid possible situations like trying to create a repository version at the same time as another task is creating a repo version.

cutwater (approved):

PR #2786, opened 2021-09-08
Pin setuptools in docker base image

awcrosby:
Attempt to fix build server conflict with use_2to3, may add pyproject.toml if also necessary.


drodowic (approved):
lgtm

PR #3226, opened 2023-08-21
docker-compose is no more needed on CI

rochacbruno:
GHA image comes with compose v2 installed


himdel (approved):
LGTM :+1: 

Fixes the CI error seen in https://github.com/ansible/galaxy/actions/runs/5859314696/job/15885052496?pr=3223#step:4:20

And the env should already have "Docker Compose v1 1.29.2" by default (https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md).

PR #3419, opened 2024-11-20
Adding community team members to namespace request issues

traytorous:
I removed Chyna from this list because she is switching from P.E and I added the community team memeber to the list because of the Team merger.


traytorous:
If everthing is good here do not merge untill tommorow

PR #3521, opened 2025-07-17
Don't assign Namespace assigns

gundalow:
Don't auto assign Galaxy namespace requests.
Slack notifies us of work.
The ticket owner MUST assign the ticket to themselves when they start working on it.


anweshadas:
Thank you @gundalow this would be helpful.

